PR #3, opened 2025-05-15
Add plonky2 benchmarks

alxkzmn:


duguorong009 (commented):
I have some comments. @alxkzmn 
- Do we need `LICENSE` file included in the repo?
- In the `Cargo.toml` file, I see that you use the custom fork of `plonky2` repo.
Can we just use the original repo? 
Or can we note that we use the custom fork, in `README.md`, if there is anything special in custom fork?

alxkzmn:
@duguorong009 
- yes, I think we also have to put any licenses and copyrights in every corresponding folder, if the original repo contained one
- good catch, I have switched to the published `plonky2` crate, however, we still need my fork for `plonky2_u32`. I believe the main reason was that the original `plonky2_u32` depended on an old version of `plonky2` and did not support serialization, so it would be impossible to measure PK/VK/proof size. I have submitted a PR for `plonky2_u32` but they haven't responded so far.

duguorong009:
Thanks for reply! @alxkzmn 
- Ok, if you think so. I'll also check my commits for missing files if any
- Hmm, it's weird that they don't respond to PR. 
Anyway, I think it is best to leave notes, in the `README.md`.
We can return when original `plonky2` has needed feature.

alxkzmn:
@duguorong009  I've added a readme note.

duguorong009 (approved):

PR #4, opened 2025-05-15
Add Plonky3 (SP1) benchmarks

alxkzmn:


plonky3-sp1/program/src/main.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    let input = &[5u8; 2048];
+    let a = sha2(input);
+
+    // Encode the public values of the program.
+    //let bytes = PublicValuesStruct::abi_encode(&PublicValuesStruct { n, a });
+
+    // Commit to the public values of the program. The final proof will have a commitment to all the
+    // bytes that were committed to.
+    //sp1_zkvm::io::commit_slice(&bytes);
```
duguorong009, line 25: I have one question.
Current code includes only the `sha256` hash function, NOT the final proof produce.
I think we should include whole process of producing final proof here.
I mean the proving time should include the time for whole process of producing proof.
Why do we omit this process?

alxkzmn: This is intentional. I think we shouldn't benchmark the VM-specific code, but try to isolate the sha256 as much as possible. This is because our goal is to benchmark plonky3 but not the sp1. What do you think?

duguorong009: Hmm, I think it makes sense.
I missed the point that we are aiming the `plonky3` for benchmarking.

duguorong009 (approved):

PR #5, opened 2025-05-15
Add Plonky3 (powdr) benchmark

alxkzmn:


plonky3-powdr/guest/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+fn sha2(input: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    hasher.update(input);
+    let result = hasher.finalize();
+    Into::<[u8; 32]>::into(result)
+}
+
+fn main() {
+    // Any serde-deserializable type can be read from a channel.
+    //let msg: Vec<u8> = read(1);
```
duguorong009: can we remove this?

plonky3-powdr/guest/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+    let msg = &[5u8; 2048];
+    let _ = sha2(msg);
+
+    //commit::commit(res);
```
duguorong009: Why should we omit this?

duguorong009 (commented):
I have one more comment. 
Can you add more details(reference, more explanation) in `README.md`?

plonky3-powdr/guest/src/main.rs
alxkzmn: This is for the same reason as with SP1 - to reduce the VM overhead and test only the SHA256

duguorong009 (approved):

PR #12, opened 2025-07-16
Add RISC Zero(zkVM) benchmarks

guorong009:


risc0/.gitignore
```diff
@@ -0,0 +1,18 @@
+*~
+*.swp
+*.swo
+.bazelrc.local
+.cache
+.devcontainer/
+.dockerignore
+.DS_Store
+.idea/
+.vim/
```
alxkzmn: Let's put those things that are not specific to RISC0 into the root gitignore, for example, .DS_Store

alxkzmn (approved):
LGTM

PR #14, opened 2025-07-21
Add jolt (zkvm) benchmark

guorong009:
## Changes
- add `utils` dir for common utilities for zkVM benchmarks
- add `guests` dir for sharing guest(e.g. sha256) between zkvm benchmarks
- add jolt zkvm benchmark for sha256


utils/src/profile.rs
```diff
@@ -0,0 +1,46 @@
...
+        profile.encode(&mut content)?;
+        file.write_all(&content)?;
+
+        println!("Profile data written to {}", file_path);
+    };
+
+    Ok(())
+}
+
+pub fn init_trace() {
```
alxkzmn: This function seems to be unused

utils/src/profile.rs
```diff
@@ -0,0 +1,46 @@
+use pprof::protos::Message;
+use std::any::Any;
+use std::fs::File;
+use std::io::Write;
+use tracing_chrome::ChromeLayerBuilder;
+use tracing_subscriber::{self, prelude::*};
+
+pub fn profile_func<F>(func: F, file_path: &str) -> Result<(), Box<dyn std::error::Error>>
```
alxkzmn: This function also seems to be unused

utils/src/bench.rs
```diff
@@ -0,0 +1,134 @@
...
+    (result, peak.load(Ordering::Relaxed))
+}
+
+#[serde_as]
+#[derive(Serialize, Tabled)]
+pub struct Metrics {
+    pub size: usize,
+    #[serde_as(as = "DurationNanoSeconds")]
+    #[tabled(display_with = "display_duration")]
+    pub exec_duration: Duration,
```
alxkzmn: I think it's ok to omit the total execution duration, we're interested in prover and verifier times separately.

utils/src/bench.rs
```diff
@@ -0,0 +1,134 @@
...
+    stop.store(true, Ordering::Relaxed);
+    monitor.join().unwrap();
+
+    (result, peak.load(Ordering::Relaxed))
+}
+
+#[serde_as]
+#[derive(Serialize, Tabled)]
+pub struct Metrics {
+    pub size: usize,
```
alxkzmn: `size` appears to be unused, and the name is also not very informative. I guess we will need to report proof size and the has input size (or some other kind of input size), but seems like `proof_bytes` is meant for the proof size (although I'd prefer to call it proof_size).

utils/src/bench.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    stop.store(true, Ordering::Relaxed);
+    monitor.join().unwrap();
+
+    (result, peak.load(Ordering::Relaxed))
+}
+
+#[serde_as]
+#[derive(Serialize, Tabled)]
+pub struct Metrics {
```
alxkzmn, line 59: Should we do some composition, like having BaseMetrics struct and then include it into custom structs for Hash and ECDSA (hash will have input size and hash function name but ECDSA won't)? Or just a fat struct that has all fields for all kinds of benches?

utils/src/profile.rs
guorong009: Removed in b6fa599

utils/src/profile.rs
guorong009: Removed in b6fa599

utils/src/bench.rs
guorong009: Removed in f4658a7. Also, removed the `cycles` field.

utils/src/bench.rs
guorong009: Rename the `size` as `input_size(bytes)` & `proof_bytes` as `proof_size(bytes)` in f4658a7

utils/src/bench.rs
guorong009: I think we don't need such composition.
In `kkrt-labs-benchmarks`, they just use `Metrics` for everything. In case of ECDSA, the `input_size` is `1`.
Maybe, such struct is needed when we wanna compose final table from every individual metrics.
If we really need such structs, I think the fat structs are good enough. e.g., `HashMetrics`, `ECDSAMetrics`
They share most of fields, only few differ. Also, we wouldn't need many of them, but only 2 or 3 structs.

utils/src/bench.rs
alxkzmn: Btw we may include some VM-specific metrics like cycles, let's leave that for the future.

guorong009: ok, roll back `cycles` field in 2d269cf

alxkzmn (approved):

brech1 (approved):

PR #17, opened 2025-07-24
Restore the Binius RAM measurement code

alxkzmn:
The Binius RAM measurement code was lost when grouping the separate ZKP systems into this single repository. Restoring the code **_for the reproducibility of the original results_** in https://pse.dev/blog/efficient-client-side-proving-for-zkid#introduction


guorong009 (approved):
LGTM!

PR #24, opened 2025-08-06
feat: implement risc0 ere based benchmarks

brech1:
# Description

This PR introduces a new RISC0 ere-based benchmark for SHA-256 and refactors with compatible changes the existing Provekit suite.

## Changes

- Added a RISC0 guest and host runner for SHA-256 benchmarking.
- Added `.csv` to `.gitignore` to exclude benchmark result files.
- Updated the memory usage calculation in the benchmark utils.
- Fixes current risc0 crate import issue.

## Results

Initial benchmark results for the new RISC0 SHA-256 implementation are included below:

| input_size | proof_duration | verify_duration | cycles | proof_size | peak_memory |
| :--------- | :------------- | :-------------- | :----- | :--------- | :---------- |
| 256B       | 12.36s         | 13.4ms          | 27.7K  | 218KB      | 732.8MB     |
| 512B       | 24.29s         | 13ms            | 50.9K  | 218KB      | 1.09GB      |
| 1KB        | 24.41s         | 13ms            | 97.5K  | 218KB      | 1.13GB      |
| 2KB        | 33.14s         | 13ms            | 190.6K | 218KB      | 2.34GB      |
| 4KB        | 1:06           | 13.2ms          | 376.8K | 218KB      | 4.44GB      |

### DIFF
--- a/ere/guests/sha256/Cargo.toml
+++ b/ere/guests/sha256/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "risc0-sha256-guest"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+risc0-zkvm = { version = "2.2", default-features = false, features = [
+    "std",
+    "unstable",
+] }
+sha2 = { git = "https://github.com/risc0/RustCrypto-hashes", tag = "sha2-v0.10.6-risczero.0" }
--- a/ere/risc0/src/lib.rs
+++ b/ere/risc0/src/lib.rs
@@ -1,0 +1,79 @@
+use ere_risc0::{EreRisc0, RV32_IM_RISC0_ZKVM_ELF, Risc0Program};
+use rand::RngCore;
+use std::collections::HashMap;
+use std::path::Path;
+use std::time::Instant;
+use utils::bench::{Metrics, benchmark};
+use zkvm_interface::{Compiler, Input, ProverResourceType, zkVM};
+
+const INPUT_EXP: [u8; 5] = [8, 9, 10, 11, 12];
+const CSV_OUTPUT: &str = "risc0_sha256.csv";
+
+/// Risc0 SHA-256 benchmark.
+struct Sha256Benchmark {
+    program: Risc0Program,
+    inputs: HashMap<u8, (usize, &'static [u8])>,
+}
+
+impl Sha256Benchmark {
+    /// Compiles the guest program and generates all necessary inputs.
+    fn new(exponents: &[u8]) -> Self {
+        let guest_relative = Path::new("../guests/sha256");
+        let program = RV32_IM_RISC0_ZKVM_ELF
+            .compile(guest_relative)
+            .expect("Failed to compile guest program");
+
+        let inputs = exponents
+            .iter()
+            .map(|&exp| {
+                let size = 1usize << exp;
+                let mut data = vec![0u8; size];
+                rand::thread_rng().fill_bytes(&mut data);
+                let static_data: &'static [u8] = Box::leak(data.into_boxed_slice());
+                (exp, (size, static_data))
+            })
+            .collect();
+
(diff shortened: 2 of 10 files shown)


ere/guests/sha256/Cargo.toml
alxkzmn, line 1: We already have a top-level `guests` crate. We need to decide between that and having guests inside `ere`.

ere/risc0/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+use ere_risc0::{EreRisc0, RV32_IM_RISC0_ZKVM_ELF, Risc0Program};
+use rand::RngCore;
+use std::collections::HashMap;
+use std::path::Path;
+use std::time::Instant;
+use utils::bench::{Metrics, benchmark};
+use zkvm_interface::{Compiler, Input, ProverResourceType, zkVM};
+
+const INPUT_EXP: [u8; 5] = [8, 9, 10, 11, 12];
```
alxkzmn, line 9: These input parameters are repeated across different benchmarks, we probably need to put them in one place, maybe `utils`.

utils/src/bench.rs
```diff
@@ -1,28 +1,34 @@
...
     unsafe {
-        let mut out: libc::rusage = std::mem::zeroed();
-        libc::getrusage(libc::RUSAGE_SELF, &mut out);
+        let mut self_usage: libc::rusage = std::mem::zeroed();
+        libc::getrusage(libc::RUSAGE_SELF, &mut self_usage);
+
+        let mut child_usage: libc::rusage = std::mem::zeroed();
+        libc::getrusage(libc::RUSAGE_CHILDREN, &mut child_usage);
+
+        let total_maxrss = self_usage.ru_maxrss + child_usage.ru_maxrss;
```
alxkzmn, line 23: 👏 👏 👏 Thanks for fixing this!

ere/guests/sha256/Cargo.toml
guorong009: If `ere` can be used for all of upcoming zkVMs, we can remove top-level `guests` and just use the `ere/guests`.
If not, we should decide which one is best, I think.

ere/risc0/src/lib.rs
guorong009: Yes, I agree. 
BTW, there is `utils::metadata::SHA2_INPUTS` constant, which can be used like `INPUT_EXP`.
Atm, it is just `SHA2_INPUTS: [usize; 1] = [2048]`.
https://github.com/privacy-scaling-explorations/csp-benchmarks/blob/CSP-Q3-2025/utils/src/metadata.rs

ere/guests/sha256/Cargo.toml
brech1: I agree, but would like to add further restructuring changes in future PR since having a single zkVM implementation wasn't enough to justify it.

ere/risc0/src/lib.rs
brech1: I wouldn't do this until we're willing to implement multiple input sizes for all sha256 benchmarks, so we can work on a broader abstraction. Same can be applied to input generation.

brech1:
I removed provekit update from the PR

ere/guests/sha256/Cargo.toml
guorong009: One more info. 
The top-level `guests` is only used for `jolt` zkVM at the moment.
If `jolt` would be replaced by `ere` in the future, we can just remove the top-level `guests` crate.

ere/risc0/src/lib.rs
alxkzmn: We will 100% run them all with varying input sizes, but ok if you want to do it in a separate task.

alxkzmn (approved):

guorong009 (approved):

PR #25, opened 2025-08-07
feat: use `criterion` for benchmarks

guorong009:
## Description
Use the `criterion` for benchmarks 

## Changes
- refactor the `provekit` benchmarks for use of `criterion`


alxkzmn (dismissed):
LGTM!

brech1:
@duguorong009 I submitted my updates, removed the binary file and implemented display for the `SubMetrics`. Everything should run correctly on `cd provekit/ && cargo bench`

provekit/src/lib.rs
```diff
@@ -1,138 +1,107 @@
...
-        let toml_path = format!("{}/Prover.toml", circuit_dir);
-
-        generate_prover_toml(&toml_path, &input_file, size)?;
+    /// Runs the proving algorithm.
+    pub fn run_prove(&self, input_exp: u32) -> NoirProof {
+        let proof_scheme = self.proof_schemes.get(&input_exp).unwrap();
+        let toml_path = self.toml_paths.get(&input_exp).unwrap();
+        let witness_map = proof_scheme
+            .read_witness(toml_path.to_str().unwrap())
+            .expect("Failed to read witness");
```
brech1, line 85: This could impact performance of the proving, but imo is negligible. Can be removed, just need to add the `noirc_abi` crate to use the `InputMap` type

guorong009: I think it is ok.

alxkzmn: We need to bench the witness generation separately, let's do it in a separate task.

alxkzmn (dismissed):

brech1 (approved):

alxkzmn (approved):

PR #28, opened 2025-08-12
ref: merge criterion bench and the measure of other metrics

guorong009:
## Description
Run the criterion bench and submetrics(RAM, proof_size, etc) measurement in one command

## Changes
Refactor the following benchmarks to apply above idea:
- binius
- plonky2
- plonky3-powdr
- plonky3-sp1

### DIFF
--- a/binius/benches/sha256_bench.rs
+++ b/binius/benches/sha256_bench.rs
@@ -1,9 +1,26 @@
 // Copyright 2024-2025 Irreducible Inc.
 
 use binius::bench::{prove, sha256_no_lookup_prepare, sha256_with_lookup_prepare, verify};
+use binius_utils::SerializeBytes;
 use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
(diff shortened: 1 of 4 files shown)


binius/benches/sha256_bench.rs
```diff
@@ -1,9 +1,26 @@
...
 use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use utils::{
+    bench::{SubMetrics, display_submetrics, measure_peak_memory, write_json_submetrics},
+    metadata::SHA2_INPUTS,
+};
 
 fn sha256_no_lookup(c: &mut Criterion) {
+    let mut all_metrics = Vec::new();
+
+    for &num_byte in SHA2_INPUTS.iter() {
```
alxkzmn, line 14: let's call it `input_size` instead of num bytes across all the code

guorong009: Done af27725

PR #31, opened 2025-08-18
feat: measure the RAM footprint of benchmarks by using `/usr/bin/time` util

guorong009:
## Description
- Add the shell script for using the `/usr/bin/time` util in measuring the RAM footprint of benchmarks
- Refactor the relating benchmarks

## Changes
- add `measure_mem_avg.sh` script
  - it can set number of runs(e.g `NUM_RUNS=4 ../measure_mem_avg.sh ...`; default value: 10)
  - it can also set the output json file(e.g `../measure_mem_avg.sh --json sha2_binius_mem_report.json -- [CMD]`)
- refactor the `binius`, `plonky2`, `plonky3-powdr`, `provekit` benchmarks
  - move the `SubMetrics` measurement to criterion bench code
  - refactor the `provekit` benchmark

## NOTE
- omit the `plonky3-sp1` since it could be replaced by `ere`


alxkzmn:
@guorong009 I see that you retained the RAM measurement in the old "measure" binaries, but I think it's ok to delete them and move the proof/preprocessing size measurement into the criterion benchmarks before the timings measurement.

alxkzmn (approved):

brech1 (approved):

PR #32, opened 2025-08-19
feat: miden ere based benchmarks

### LINKED ISSUE #34: Provide GITHUB_TOKEN higher up in the workflow
Seems like we need to move 
```yaml
env:
          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
```
somewhere on the top level because Noir has that build script that needs the token and it's called in places where we don't always expect (like clippy).

brech1:
# Description

This PR introduces a new miden ere-based benchmark for SHA-256 and refactors with compatible changes the existing Provekit suite.

## Changes

- Added miden benchmarks
- Fixed risc0 benchmarks

## Related Issues

- Resolves #34 

## Results

| input_size | proof_duration | verify_duration | cycles | proof_size | peak_memory |
|---|---|---|---|---|---|
| 256B | 2.21s | 1.1ms | 42.5K | 79.3KB | 989.4MB |
| 512B | 4.49s | 842.2µs | 76.2K | 83.2KB | 1.91GB |
| 1KB | 10.54s | 1ms | 143.6K | 91.7KB | 3.16GB |
| 2KB | 20.94s | 1ms | 278.4K | 96.7KB | 4.45GB |
| 4KB | 45.42s | 1ms | 548.1K | 102.9KB | 9.07GB |


ere/miden/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+use ere_miden::{EreMiden, MidenTarget};
+use rand::RngCore;
+use std::collections::HashMap;
+use std::path::Path;
+use std::time::Instant;
+use utils::bench::{Metrics, benchmark};
+use zkvm_interface::{Compiler, InputItem, ProverResourceType, zkVM};
+
+const INPUT_SIZES: [usize; 5] = [256, 512, 1024, 2048, 4096];
```
alxkzmn: Should use SHA2_INPUTS from utils/src/metadata.rs

ere/miden/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+        zkvm.verify(&proof).unwrap();
+        let verify_duration = verify_start.elapsed();
+
+        let mut metrics = Metrics::new(size);
+        metrics.proof_duration = proof_duration;
+        metrics.verify_duration = verify_duration;
+        metrics.cycles = execution_report.total_num_cycles;
+        metrics.proof_size = proof.len();
+        metrics
+    }
```
alxkzmn, line 63: Looks like this is very similar to the RISC0 code, can it be moved to a shared trait and reused?

ere/risc0/src/lib.rs
```diff
@@ -6,45 +6,41 @@ use std::time::Instant;
 use utils::bench::{Metrics, benchmark};
 use zkvm_interface::{Compiler, Input, ProverResourceType, zkVM};
 
-const INPUT_EXP: [u8; 5] = [8, 9, 10, 11, 12];
+const INPUT_SIZES: [usize; 5] = [256, 512, 1024, 2048, 4096];
```
alxkzmn: Should use SHA2_INPUTS from utils/src/metadata.rs

alxkzmn (changes requested):
Please review my comments, I think there's potential for more DRY.

ere/miden/src/lib.rs
brech1: It's a good idea, however I should try this in the next PR when I migrate all zkVMs to `ere` just in case I'm missing something, just two at the moment.

alxkzmn: Created an issue: https://github.com/privacy-scaling-explorations/csp-benchmarks/issues/35

alxkzmn (approved):

Issue #29, opened 2025-08-13
Separate GitHub workflows


alxkzmn:
The following workflows should run on pull request:
 - Clippy;
 - Fmt.
  
The following workflows should run on a manual run and on schedule (e.g., at 10pm GMT):
- Time benchmarks;
- RAM benchmarks.

PR #36, opened 2025-08-21
ci: separate noir installation github action

### LINKED ISSUE #27: Separate Noir installation github action
The following code is repeating in the workflow file:
```
- name: Install Nargo
        uses: noir-lang/noirup@v0.1.4
        with:
          toolchain: stable
```
It can be separated into an action and reused.

guorong009:
- Closes #27  
- Partially related to #29 (split the `fmt` and `clippy` to separate workflow)

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -50,10 +50,13 @@ jobs:
       - name: Install nightly-2024-08-01 with rust-src for Powdr
         run: rustup toolchain install nightly-2024-08-01 --component rust-src
       
-      - name: Build full workspace (release)
+      - name: Install Nargo
+        uses: ./.github/actions/install-nargo
+      
+      - name: Build full workspace
         env:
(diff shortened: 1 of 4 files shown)


alxkzmn (approved):
LGTM

.github/workflows/lints.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Install nightly-2024-08-01 with rust-src for Powdr
+        run: rustup toolchain install nightly-2024-08-01 --component rust-src
+      
+      - name: Install Nargo
+        uses: ./.github/actions/install-nargo
+      
+      - name: Build full workspace
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }} # for ACIR (required by protoc-prebuilt)
+        run: cargo build --workspace
```
brech1, line 59: Are these steps needed if we just want to check formatting?

guorong009: No, they are not needed for formatting, but for clippy.
Because of `SP1` needing `include_elf!(...)`, we need to build the workspace.
That's why I had to include above step.

Issue #33, opened 2025-08-20
Unify benchmark inputs across all benchmarks


alxkzmn:
The benchmarks have to use the `SHA2_INPUTS` from `utils/src/metadata.rs`

alxkzmn:
https://github.com/privacy-scaling-explorations/csp-benchmarks/pull/38/files#diff-db427605508c51a8831fdbfb49102aade6fe9a8175922290fac2a24c5031514bL10

alxkzmn:
Please note that there may be bugs similar to this one: https://github.com/privacy-scaling-explorations/csp-benchmarks/blob/1d563d582fd5f9ecc7976de6346ff9dca70bd320/provekit/src/bin/measure_mem.rs#L7 - looks ok, but if measured by the RAM measurement script it will measure the loop across all input sizes, not each input size separately.

PR #37, opened 2025-08-21
restructure: provekit inputs

brech1:
# Description

Updates provekit to use `SHA256_INPUTS` instead of exponents for the benchmarks.

# Related Issues

- Partially solves #33


guorong009 (approved):
LGTM!

PR #38, opened 2025-08-28
feat: collect benchmarks(0)

guorong009:
## Description
- Standardize the JSON files(criterion, memory report, metrics) with naming
  - naming: `[target]_[size]_[proving_system]_[feat]_{...}.json`
    Example: `sha256_2048_binius_no_lookup_prove.json`, `sha256_2048_provekit_mem_report.json`, `sha256_2048_powdr_metrics.json`
- Collect benchmarks of proving systems from 3 types of json files - criterion(time), memory report, metrics

## Changes
- update the namings of JSON files in criterion code, README of some proving systems
  - covers: `binius`, `plonky2`, `plonky3-powdr`, `provekit`
- implement the logic of collecting benchmarks in `utils` crate
  - add bin code of collecting benchmarks
- remove unused code(related to ECDSA) from `utils` 

## NOTE  
- This PR covers only few proving systems, not all. (`sp1` is omitted since would be moved to `ere`)
- The zkVMs and others(`stwo`, `expander`) should be integrated in future PRs.


guorong009:
[collected_benchmarks.json](https://github.com/user-attachments/files/22029327/collected_benchmarks.json)
This is the final JSON file we get from collecting the benchmarks.

utils/src/bin/collect_benchmarks.rs
```diff
@@ -0,0 +1,168 @@
...
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+use std::{fs, io};
+use utils::bench::CollectedMetrics;
+
+fn main() -> io::Result<()> {
+    let benchmark_target = "sha256".to_string();
+    let benchmark_input_size = 2048;
+
+    let mut proving_systems: HashMap<String, Vec<String>> = HashMap::new();
```
alxkzmn: I think we can make the system more robust and friendly to future contributions if we just collect all JSON files in the expected directory, and expect each JSON file to have a certain format. Skip the file if it can't be deserialized into the expected struct. The struct should contain the bench identifier/name and features, so instead of having separate SubMetrics and CollectedMetrics, we have just one kind of metrics. We simply ask future contributors to write out these metrics to a JSON file in a directory where we expect it, this way they won't forget to assign a name and "features", if any. This way we won't need to maintain things on two sides - in the proving system crate and here in the collection script.

guorong009: Good point!
I'll try to refactor the codebase as recommended!

utils/src/bench.rs
```diff
@@ -176,3 +176,42 @@ pub fn write_json_submetrics(output_path: &str, metrics: &SubMetrics) {
...
+        CollectedMetrics {
+            name,
+            feat,
+            target,
+            input_size,
+            proof_size: 0,
+            preprocessing_size: 0,
+            proof_duration: Duration::default(),
+            verify_duration: Duration::default(),
+            cycles: 0,
```
brech1: We should have a call regarding Metrics and the utils crate. I think it should be restructured since I'll have my own leaner crate on the zkvm side. Cycles here shouldn't be a prop! And other zkvm related code should be removed as well.

guorong009: Ok, I agree that we should have a call.
For the moment, I'll remove the zkvm related stuff from PR.

guorong009:
[collected_benchmarks.json](https://github.com/user-attachments/files/22048918/collected_benchmarks.json)
This is the latest JSON file, obtained by new collection logic.
Pls review the PR, again. @alxkzmn @brech1

utils/src/bench.rs
```diff
@@ -137,42 +137,51 @@ pub fn write_csv(out_path: &str, results: &[Metrics]) {
 }
 
 #[serde_as]
-#[derive(Serialize, Tabled, Clone, Copy)]
-pub struct SubMetrics {
+#[derive(Serialize, Deserialize, Tabled, Clone)]
+pub struct Metrics1 {
```
alxkzmn: Let's remove "1" from the name (and also from the corresponding methods name like write_json_metrics1 and display_metrics1)

utils/src/bin/collect_benchmarks.rs
```diff
@@ -0,0 +1,132 @@
+use glob::glob;
+use serde_json::Value;
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+use std::{fs, io};
+use utils::bench::Metrics1;
+
+fn main() -> io::Result<()> {
+    let mut benchmarks: HashMap<String, Metrics1> = HashMap::new();
```
alxkzmn: I think this can be simplified to an array (vector?), this way the frontend UI can read a uniform JSON array element by element instead of iterating over the map keys.

alxkzmn:
Please create a contribution guide that defines the naming rules for criterion functions, RAM measurement output, etc.

utils/src/bench.rs
guorong009: The reason of using `Metrics1` is that there is another `Metrics`(which is for zkVM) in the same file. 
After we move zkVM-related `Metrics` to somewhere else, I can name `Metrics1` back to `Metrics`.

utils/src/bin/collect_benchmarks.rs
guorong009: Done 379dc2c

CONTRIBUTING.md
alxkzmn, line 1: I think we need to be a bit more clear on what we expect. We expect the benchmarking code crate to be added to the workspace (can link to the top-level Cargo.toml here to make it clear). Our code that collects the results expects 3 JSON files in a certain place and with a certain naming convention (as you explained). Can give the example of how we are writing the files by serializing the Metrics struct by linking to the existing code, also need to mention that we have already implemented the RAM measurement script that the contributors can use. BTW I think we need to fix the runs to 10 and do not write that number to the output file so that the contributors don't have to think about how many runs they should do.

utils/src/bench.rs
guorong009: Done 530aac7

guorong009:
[collected_benchmarks.json](https://github.com/user-attachments/files/22080139/collected_benchmarks.json)
(This is the JSON file of benchmarks collected)
@alxkzmn @brech1 
I've refactored the codebase as discussed.
Please take a look again.

alxkzmn (approved):
LGTM!

brech1 (approved):

PR #39, opened 2025-09-03
ref: unify inputs across the benchmarks

### LINKED ISSUE #33: Unify benchmark inputs across all benchmarks

guorong009:
## Description
Use the `utils::metadata::SHA2_INPUTS` for every benchmarks

## Related issues
Closes #33 

## Changes
- refactor the several benchmarks so that they use above 
  - `binius`, `plonky2`, `plonky3-powdr`, `provekit`
- fix the `provekit` ci failure

### DIFF
--- a/binius/benches/sha256_bench.rs
+++ b/binius/benches/sha256_bench.rs
@@ -9,93 +9,95 @@ use utils::{
 };
 
 fn sha256_no_lookup(c: &mut Criterion) {
-    // Measure the metrics
-    let input_size = SHA2_INPUTS[0];
-    let metrics = sha256_binius_no_lookup_metrics(input_size);
+    for input_size in SHA2_INPUTS {
+        // Measure the metrics
+        let metrics = sha256_binius_no_lookup_metrics(input_size);
 
-    let json_file = format!("sha256_{input_size}_binius_no_lookup_metrics.json");
-    write_json_metrics(&json_file, &metrics);
+        let json_file = format!("sha256_{input_size}_binius_no_lookup_metrics.json");
+        write_json_metrics(&json_file, &metrics);
 
-    // Run the benchmarks
-    let mut group = c.benchmark_group(format!("sha256_{input_size}_binius_no_lookup"));
-    group.sample_size(10);
-    let allocator = bumpalo::Bump::new();
+        // Run the benchmarks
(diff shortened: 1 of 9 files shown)


binius/benches/sha256_bench.rs
```diff
@@ -3,83 +3,98 @@
...
-use utils::bench::{Metrics, write_json_metrics};
+use utils::{
+    bench::{Metrics, write_json_metrics},
+    metadata::SHA2_INPUTS,
+};
 
 fn sha256_no_lookup(c: &mut Criterion) {
     // Measure the metrics
-    let input_size = 2048;
+    let input_size = SHA2_INPUTS[0];
```
alxkzmn: Let's loop across the whole input array.

alxkzmn: Please check that across the whole codebase, so far only ProveKit time benchmarks are doing that, but we need it for all proving systems and for RAM as well.

guorong009: Add the commit(833d6e8) for looping across the input array, in criterion benches.

Regarding the RAM, we should think about how to do that in rust code using `SHA2_INPUTS` array.
The reason is that current RAM measurement is done with shell script.

Also, some benchmarks(e.g. plonky2) are the hardcoded circuits for 2048-byte inputs.
Whenever we add new input size to array, we should refactor those hardcoded benchmarks.

brech1 (approved):
Amazing!

provekit/src/bin/measure_mem.rs
```diff
@@ -1,10 +1,6 @@
 use provekit::ProvekitSha256Benchmark;
-use utils::metadata::SHA2_INPUTS;
 
 fn main() {
-    let bench_harness = ProvekitSha256Benchmark::new(&SHA2_INPUTS);
-
-    for &size in SHA2_INPUTS.iter() {
-        let _proof = bench_harness.run_prove(size);
-    }
+    let bench_harness = ProvekitSha256Benchmark::new(2048);
```
alxkzmn, line 4: I think there shouldn't be a hardcoded input here, we need to measure RAM for all input sizes.

alxkzmn (approved):

PR #40, opened 2025-09-05
restructure: zkvm benchmarks

brech1:
# Description

This PR restructures the current zkVM benchmarks. A new `zkvm` directory has been created to host them.

## Interface

A new `Benchmark` struct has been implemented to create new benchmark instances:

```rust
pub struct Benchmark<C, V>
where
    C: Compiler,
    V: zkVM,
{
    vm: V,
    vm_name: String,
    bench_name: String,
    _phantom: PhantomData<C>,
}
```

A set of traits can be found on the `traits.rs` module to deal with input generation and loading program into zkVM instances.

## Comments

- Miden benchmarks have not been included. The `ere-miden` implementation lies on a fork which is outdated respective to the current ere `zkvm-interface`, would need to merge it to `ere` and then implement it here.
- Utils metrics haven't been modified.


zkvm/benches/sha256.rs
```diff
@@ -0,0 +1,91 @@
...
+        results.push(output.1);
+    }
+
+    write_csv("jolt-sha256.csv", &results);
+}
+
+fn main() {
+    // risc0_sha256();
+    // sp1_sha256();
+    jolt_sha256();
```
alxkzmn: These should be a single struct+trait, they only differ in parameters of `Benchmark`. This part
```rust
let configs = SHA2_INPUTS.map(|size| Sha256Config::new(size));

    let generator = Sha256Generator;
    let benchmark = Benchmark::new(
        &RV32_IM_RISC0_ZKVM_ELF,
        "risc0",
        Sha256::NAME,
        &Risc0Builder,
    )
    .unwrap();
```
 can be in `::new(<elf, other bench params>)`
and the rest can be in a trait method, like `run_benchmark()`.

brech1: Yeah nice, I didn't know if the memory measurements were good enough as they are here, since it could actually be integrated inside the `Benchmark::run`. But can go ahead and do this.

.github/workflows/zkvm_benchmarks.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: cargo fmt --all -- --check
+
+      - name: Build
+        run: cargo build --release
+
+      - name: Run clippy
+        run: cargo clippy --all-targets --all-features -- -D warnings
+
+      - name: Run benchmarks
+        run: cargo bench
```
alxkzmn, line 75: Is there a way to parallelize them?

zkvm/benches/sha256.rs
alxkzmn, line 1: It's possible to refactor this to not repeat the code - the method can take the ELF, id, name and builder as args.

zkvm/src/programs/sha256.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl InputBuilder<Sha256> for EreRisc0 {
+    type Data = Vec<u8>;
+
+    fn build_input(data: Self::Data) -> Input {
+        let static_data: &'static [u8] = Box::leak(data.into_boxed_slice());
+        let mut input = Input::new();
+        input.write(static_data);
+        input
+    }
```
alxkzmn: Why static and leak? It can simply be
```suggestion
fn build_input(data: Self::Data) -> Input {
        let mut input = Input::new();
        input.write(data);
        input
    }
```

alxkzmn: Btw, found a similar thing in `ere/miden/src/lib.rs`, could also be done without static and leaking.

zkvm/guests/jolt/sha256/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+#![cfg_attr(feature = "guest", no_std)]
+
+use sha2::{Digest, Sha256};
+
+#[jolt::provable]
+fn sha2(_input: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    // Ere is not handling program inputs for jolt at the moment.
+    // For now, I'll hardcode a 2048 byte input.
```
alxkzmn, line 9: Will we be able to fix it in the future? Do we need to track it in an issue?

alxkzmn (changes requested):
I believe this could be refactored: https://github.com/privacy-ethereum/csp-benchmarks/pull/40/files#r2332012499
Would appreciate if you address other comments as well!

zkvm/guests/jolt/sha256/src/lib.rs
brech1: Not an issue per se just an implementation remark, there are other ways to adapt this for multiple inputs.

.github/workflows/zkvm_benchmarks.yml
brech1: Yes. It's possible to do so by vm and/or program. Not with the current setup however, and will change with `usr/bin/time`

brech1:
Performed requested changes

zkvm/src/programs/sha256.rs
```diff
@@ -0,0 +1,62 @@
...
+            fn build_input(data: Self::Data) -> Input {
+                let mut input = Input::new();
+                input.write(data);
+                input
+            }
+        }
+    };
+}
+
+impl_default_input_builder!(EreRisc0);
```
alxkzmn: The name could be more informative, like `impl_byte_vec_input_builder`

alxkzmn (dismissed):

zkvm/src/programs/sha256.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    fn generate(&self, config: &Sha256Config) -> (Self::Data, usize) {
+        let mut rng = StdRng::seed_from_u64(1337);
+        let mut data = vec![0; config.input_size];
+
+        rng.fill_bytes(&mut data);
+
+        (data, config.input_size)
+    }
+}
```
alxkzmn: Parameterizing by both program and benchmark config seems unnecessary, won't only one of them be enough?  Right now you're only using the Config to get the input size, and you're implementing it for a particular kind of generator, so you probably know which program it is for.

guorong009 (dismissed):

zkvm/src/programs/sha256.rs
brech1: Thanks, I removed the program generic

alxkzmn (dismissed):
LGTM! I'll resolve the conflict and merge

PR #42, opened 2025-09-09
Fix Polyhedra Expander benchmark

alxkzmn:
**Key changes:**

* Created a SHA-256 circuit that is using Mersenne31 field without SIMD (doesn't perform parallel proving, runs a single SHA-256 circuit);
* Added the crate to the workspace;
* Implemented benchmarking in the same way as in the other crates, including metrics reporting;
* Had to add a `build.rs` script that generates circuit code for multiple input sizes at build time because Expander generates circuits using a macro and the input size needs to be known at compile time.


utils/src/metadata.rs
```diff
@@ -1 +1 @@
-pub const SHA2_INPUTS: [usize; 1] = [2048];
+pub const SHA2_INPUTS: [usize; 2] = [1024, 2048];
```
guorong009: Can we add new `input_size` later?
I think this can cause bench failure in other benchmarks, or wrong benchmarks.
For example, `provekit` now has only 2048-byte input_size circuit.
This sudden addition can bring the failure in benchmarking.
Also, `plonky2` only has 2048-byte input_size handling.
It would bring wrong benchmarking for 1024-byte input_size.
How about using hardcoded value(1024) for polyhedra-expander temporarily?

guorong009 (commented):
I see that dependencies are still not specific - they are fixed by either git branch or repository.
Didn't the expander team still fix the version or git tagging in their repositories?

utils/src/metadata.rs
alxkzmn: I have only added this to test that the macro generates the circuits correctly, I will roll it back but we need to start working with real variable inputs ASAP.

alxkzmn:
> I see that dependencies are still not specific - they are fixed by either git branch or repository. Didn't the expander team still fix the version or git tagging in their repositories?

good point, I'll pin the revisions!

utils/src/metadata.rs
guorong009: Yes, I agree.

alxkzmn:
@guorong009 unfortunately I could neither pin the revisions nor patch the internal dependencies. In the former case only the direct dependencies are pinned and the compiler sees version mismatch between the pinned revision and internal deps (that are still pointing to branch=main). In the latter case the compiler says that the patch is "pointing to the same source" because the pinned revision is the same as the current branch=main.  I will let them know our concerns, but for now let's rely on the lockfile, the revision is specified there explicitly.

guorong009 (approved):
LGTM! Great work! 🚀

brech1 (approved):

PR #53, opened 2025-09-10
ref: improve the process of memory measurement

### LINKED ISSUE #46: Run SHA-256 RAM benchmarks for variable input sizes

guorong009:
## Description
Improve the memory measure process by:
- isolate the memory measure into separate binary
- run the shell script in rust code
- loop through the `utils::metadata::SHA2_INPUTS` for various input size

## Related issues
Closes #46 

## Changes
Refactor the `binius`, `plonky2`, `plonky3-powdr` and `provekit` as the following:
- add the `sha256_[(optional)feat]_mem` binary(e.g. `sha256_no_lookup_mem`)
- run the memory measure shell script with above binary in the `measure_mem` binary
- add more renaming(e.g. all the memory measure binary is called `measure_mem`)
- update README


binius/src/bin/sha256_with_lookup_mem.rs
```diff
@@ -0,0 +1,26 @@
...
+fn main() {
+    let args = Args::parse();
+
+    match args.input_size {
+        2048 => sha256_2048_binius_with_lookup_mem(),
+        _ => panic!("Unsupported input size"),
+    }
+}
+
+fn sha256_2048_binius_with_lookup_mem() {
```
alxkzmn: I think eventually there won't be a hardcoded function per each input size. Instead, sha256_with_lookup_prepare should take input size as an argument.

binius/src/bin/measure_mem_no_lookup.rs
```diff
@@ -0,0 +1,31 @@
...
+            .arg("--json")
+            .arg(json_file)
+            .arg("--")
+            .arg(binary_path)
+            .arg("--input-size")
+            .arg(input_size.to_string())
+            .output()
+            .expect("failed to execute script");
+
+        println!("{}", String::from_utf8_lossy(&output.stdout));
```
alxkzmn: This code is repeated multiple times, it can be extracted to utils and reused across all benchmarks (all file names and binary names will be the arguments to a function).

provekit/src/bin/sha256_mem.rs
```diff
@@ -0,0 +1,23 @@
...
+    match args.input_size {
+        2048 => sha256_2048_provekit_mem(),
+        _ => panic!("Unsupported input size"),
+    }
+}
+
+fn sha256_2048_provekit_mem() {
+    let bench_harness = ProvekitSha256Benchmark::new(2048);
+    let _proof = bench_harness.run_prove();
+}
```
alxkzmn: Instead of 
```rust
match args.input_size {
        2048 => sha256_2048_provekit_mem(),
```
you can directly do 
```rust
ProvekitSha256Benchmark::new(args.input_size)
```

alxkzmn (changes requested):
We shouldn't have the input size hardcoded in any function names. Instead, size should be the argument.

binius/src/bin/sha256_with_lookup_mem.rs
guorong009: Yes, I agree with you.
The reason I do this way is that the current codes are not full ready for variable input size.

What about applying the variable input size to every benchmarks, before we come back to this PR? similar to PR #55 
We can set this PR as draft for a while.

provekit/src/bin/sha256_mem.rs
guorong009: Done 621f666

binius/src/bin/measure_mem_no_lookup.rs
guorong009: Done 21dfcc3

binius/src/bin/sha256_with_lookup_mem.rs
alxkzmn: No, we can prepare the code now in this PR and save time integrating it later. As I imagine, later we will have
```rust
fn main() {
    let args = Args::parse();
    sha256_binius_with_lookup_mem(args.input),
}
```
Even though now sha256_with_lookup_prepare doesn't take an input size, you can still pass the argument and simply ignore it for now, the argument will be used later when we upgrade to Binius64 and the variable input support is implemented. You can leave a TODO for that. This way, we avoid writing code that will be 100% deleted later.

plonky2/src/bin/sha256_no_lookup_mem.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// Input size parameter
+    #[arg(long)]
+    input_size: usize,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    match args.input_size {
+        2048 => sha256_2048_plonky2_no_lookup_mem(),
```
alxkzmn: Same here, pass the size as an argument, ignore it, leave a TODO, but don't make a size-specific function.

plonky3-powdr/src/bin/sha256_mem.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// Input size parameter
+    #[arg(long)]
+    input_size: usize,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    match args.input_size {
+        2048 => sha256_2048_powdr_mem(),
```
alxkzmn: Same here

guorong009: Done 53194fa

plonky2/src/bin/sha256_no_lookup_mem.rs
guorong009: Done 53194fa

binius/src/bin/sha256_with_lookup_mem.rs
guorong009: Done 53194fa

brech1 (dismissed):

alxkzmn (dismissed):
LGTM!

PR #62, opened 2025-09-17
feat: implement `Binius64` variable input size

### LINKED ISSUE #54: Run Binius SHA-256 time benchmarks with variable input
Upgrade to Binius64 first, then implement the variable input.
The circuit now takes message length as an argument:
https://github.com/IrreducibleOSS/binius64/blob/main/prover/examples/src/circuits/sha256.rs

guorong009:
Closes #54


binius64/benches/sha256_bench.rs
```diff
@@ -0,0 +1,109 @@
...
+fn sha256_bench(c: &mut Criterion) {
+    for input_size in SHA2_INPUTS {
+        // Measure the metrics
+        let metrics = sha256_binius64_metrics(input_size);
+
+        let json_file = format!("sha256_{input_size}_binius64_metrics.json");
+        write_json_metrics(&json_file, &metrics);
+
+        // RAM measurement
+        let sha256_binary_name = "sha256_no_lookup_mem";
```
alxkzmn: I think we can remove the "no_lookup" since now there is only one kind of circuit.

guorong009: Done c864f91

binius64/benches/sha256_bench.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub_witness_data
+        .serialize(&mut buf)
+        .expect("Failed to serialize public witness into byte array");
+
+    let non_pub_witness_data = ValuesData::from(witness.non_public());
+    non_pub_witness_data
+        .serialize(&mut buf)
+        .expect("Failed to serialize non public witness into byte array");
+
+    metrics.preprocessing_size = buf.len();
```
alxkzmn, line 89: Witness data is not "preprocessing" because it depends on the input. It won't be stored on the device between the prove runs, so it doesn't occupy storage and we don't need to measure it.

guorong009: Done 72a24da

binius64/benches/sha256_bench.rs
```diff
@@ -0,0 +1,99 @@
...
+        false,
+        "sha256".to_string(),
+        input_size,
+    );
+
+    let (_verifier, prover, witness, cs) =
+        prepare(input_size).expect("Failed to prepare sha256 circuit for prove/verify");
+
+    let mut buf: Vec<u8> = Vec::new();
+    cs.serialize(&mut buf)
```
alxkzmn, line 86: `circuit.populate_wire_witness(&mut filler))?;`
It seems like what needs to be persisted between the prover runs is the `circuit`, not the `cs`. Is it possible to serialize the whole circuit and measure its size?

binius64/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    C: PseudoCompressionFunction<Output<D>, 2>,
+    PC: ParallelPseudoCompression<Output<D::Digest>, 2>,
+{
+    let challenger = StdChallenger::default();
+
+    let mut prover_transcript = ProverTranscript::new(challenger);
+    prover.prove(witness.clone(), &mut prover_transcript)?;
+
+    let proof = prover_transcript.finalize();
+    tracing::info!("Proof size: {} KiB", proof.len() / 1024);
```
alxkzmn: If we don't have a `tracing_subscriber`, we can remove all the tracing calls (we probably don't need them anyway for benchmarking).

guorong009: Done 05697ec

binius64/benches/sha256_bench.rs
guorong009: Afaik, no. It's not possible to serialize the whole circuit and measure its size.

Also, I think `ConstraintSystem` can be used in the role of `circuit`. 
I mean it's enough to use the `cs` and measure its size.
In `binius64`, there is doc about prover binary.
They say that it is enough to save/store the `cs` and `witness` for prover running.
https://github.com/IrreducibleOSS/binius64/blob/main/prover/examples/README.md#prover-binary

alxkzmn: Can you generate a new witness by using only `cs`? Imagine having a mobile app with a Binius64 circuit inside. You will have a fixed circuit but will need to generate different witnesses each time you're running the prover. The stored binary should allow you to generate the witness.

guorong009: Ok, I see what you mean.
Then, we need to store circuit as binary or any external file, and measure its size.
As I said, there is no serde implementation for circuit. (`cs` has serde impl.)
I think we should implement it, like old `binius` - fork the `binius64` repo, implement serde for `circuit`, and use that fork for our benchmarking code. 
What do you think?

BTW, similar discussion here: https://github.com/IrreducibleOSS/binius64/pull/1139#discussion_r2341453290

alxkzmn: Let's leave it as is for now (serialize cs only) and try to join the conversation there. If they don't implement it we can submit a PR ourselves.

alxkzmn (approved):
LGTM! Let's deal with the `Circuit` serialization separately.

binius64/benches/sha256_bench.rs
alxkzmn: https://github.com/privacy-ethereum/csp-benchmarks/issues/63

brech1 (approved):

PR #81, opened 2025-09-25
feat: implement circom-sha256 benchmark

### LINKED ISSUE #68: Add Circom SHA-256 benchmark

guorong009:
## Description
Implement the `circom` sha256 benchmark

## Related issues
Close #68 

## Changes
- add the circom-sha256 circuit and its testing files


brech1 (dismissed):

circom/groth16/test_sha256_groth16.sh
```diff
@@ -0,0 +1,157 @@
...
+
+
+function normalProve() {
+  cd "$CIRCUIT_DIR"
+  avg_time 10 snarkjs groth16 prove sha256_test_0001.zkey witness.wtns proof.json public.json
+  echo "Prove time: $LAST_TIME seconds"
+  echo "Prove memory: $(bytes_to_human $LAST_MEM)"
+  proof_size=$(ls -lh proof.json | awk '{print $5}')
+  echo "Proof size: $proof_size"
+  preprocessing_size=$(du -ck sha256_test_0001.zkey witness.wtns proof.json public.json | grep total | awk '{print $1}')
```
alxkzmn: Witness is not a part of preprocessing because it is generated only before the proving and shouldn't be persisted between prover runs.

alxkzmn (changes requested):
Please implement the variable input and integrate the benchmarks into CI similarly to Ligetron (and the contribution guide).

circom/groth16/test_sha256_groth16.sh
guorong009: Done 71c77fc

brech1 (approved):

circom/benches/sha256_bench.rs
```diff
@@ -0,0 +1,45 @@
...
+    },
+    |(_witness_fn, _input_str, zkey_path), proof| {
+        circom::verify(proof.clone(), zkey_path.clone())
+    },
+    |(_witness_fn, _input_str, zkey_path)| {
+        // NOTE: We assume that the dir which includes "[circuit].zkey" also contains the files
+        //       needed for witness generation("[circuit].cpp", "[circuit].dat" files).
+        sum_file_sizes_in_the_dir(zkey_path).expect("Unable to compute preprocessing size")
+    },
+    |_proof: &CircomProof| 807 // NOTE: Assume that protocol is "rapidsnark" and curve is "bn128"
```
alxkzmn, line 23: Even if we know it's constant size, I think it's better to measure it so that even if some future update causes the size to change we don't need to worry about that.

alxkzmn: https://github.com/privacy-ethereum/csp-benchmarks/issues/84

alxkzmn (approved):
LGTM!

Issue #66, opened 2025-09-22
Report additional properties of proving systems and circuits


alxkzmn:
Output JSON format:
```json
[
 {
    "name": "binius64",
    "feat": "",
    "is_zkvm": false,
    "target": "sha256",
    "input_size": 128,
    "proof_duration": 259388125,
    "verify_duration": 26612229,
    "cycles": 0,
    "proof_size": 109808,
    "preprocessing_size": 382048,
    "peak_memory": 22298624
    ...
  },
  {
    "name": ...
    ...
  }
  ...
]
```

| Key | Type | Units | Is Optional | Description |
|-----|------|--------|--------------|--------------|
| name | string | – | No | Name of the codebase that is being benchmarked. |
| feat | string | – | Yes | A way to distinguish multiple versions of the same circuit tested for the same system (e.g., "no precompiles", "with precompiles"). |
| is_zkvm | bool | – | No | Whether the system is a zkVM (true) or not (false). |
| target | string | – | No | The kind of circuit that is benchmarked (e.g., "sha256"). |
| input_size | integer | bytes | No | Size of the circuit input used for this benchmark run. |
| proof_duration | integer | ns | No | Proof generation duration in nanoseconds. |
| witgen_duration | integer | ns | Yes | Time for witness generation (if not specified, proving time includes both witness generation and proving). |
| verify_duration | integer | ns | No | Proof verification duration in nanoseconds. |
| proof_size | integer | bytes | No | Size of the proof in bytes. |
| preprocessing_size | integer | bytes | No | Size of preprocessing artifacts in bytes. |
| peak_memory | integer | bytes | No | Peak RAM usage during the proving phase. |
| peak_memory_witgen | integer | bytes | Yes | Peak RAM during witness generation (if not specified, assumed included in “Peak RAM during proving”). |
| n_constraints | integer | – | No | Number of constraints in the circuit. |
| is_maintained | bool | – | No | Indicates if the system is actively maintained or deprecated. |
| is_zk | bool | – | No | Whether the system provides zero-knowledge properties. |
| is_audited | bool | – | No | Whether the system has undergone an external audit. |
| proving_system | string | – | No | Name of the underlying proving system. |
| field_curve | string | – | No | The finite field or elliptic curve used by the system. |
| iop | string | – | No | Type of Interactive Oracle Proof (IOP) used. |
| pcs | string | – | Yes | Polynomial Commitment Scheme (PCS) used. |
[...]

PR #82, opened 2025-09-25
Refactor Rust benchmarks harness

alxkzmn:
**_TLDR: Rust contributors will need to write 2-4x less code_**

This pull request refactors the benchmarking harnesses for several proving systems to use a unified, macro-based approach. The repetitive benchmarking, metrics, and RAM measurement logic in each backend's `benches/prove_verify.rs` (and similar) files has been replaced with calls to a new `define_benchmark_harness!` macro, which streamlines and standardizes the benchmarking process across the codebase.

Also, I've reduced the SHA-256 input size to 128 B to speed up the CI.


brech1 (approved):

utils/src/harness.rs
```diff
@@ -0,0 +1,349 @@
...
+
+#[derive(Clone, Debug)]
+pub struct BenchHarnessConfig<'a> {
+    pub target: BenchTarget,
+    pub system: ProvingSystem,
+    pub feature: Option<&'a str>,
+    pub is_zkvm: bool,
+    pub fixed_input_size: Option<usize>,
+    pub mem_binary_name: &'a str,
+}
```
guorong009, line 55: How about adding other useful attributes? like `is_zk: bool`

guorong009 (commented):
Overall, LGTM!
I left one comment about `BenchHarnessConfig`.

One question is: 
Can we keep 2048 as input_size for sha256 benchmark?
Even though CI is slow, I believe 2048 is the reasonable value for benchmarking.
On the other hand, we can run whole benchmarking CI, not every time we make PR, but once in a while(like once in month).

utils/src/harness.rs
alxkzmn: Let's deal with it here: https://github.com/privacy-ethereum/csp-benchmarks/issues/66

alxkzmn:
@guorong009 let's add bigger input sizes closer to finishing the milestone. For now it's ok to run with 128B (and it may be the starting size in the final array anyways).

guorong009 (approved):

PR #88, opened 2025-10-02
feat: implement `noir-sha256` benchmark

### LINKED ISSUE #18: Add Noir benchmarks
With [MoPro's promising numbers with Noir](https://zkmopro.org/docs/performance/#noir) and Noir's [popularity across other identity projects](https://github.com/noir-lang/awesome-noir/?tab=readme-ov-file#identity), it's worthwhile extending benchmarks to cover Noir for zkID.

guorong009:
Resolve #18


.github/workflows/sh_benchmarks_parallel.yml
```diff
@@ -99,12 +100,24 @@ jobs:
...
+          brew install bash
+
       - name: Setup & build Ligetron (only when folder is ligetron)
         if: ${{ matrix.folder == 'ligetron' }}
         uses: ./.github/actions/install-ligero
         with:
           ligero_path: ./ligetron/ligero-prover
           install_prefix: ${{ github.workspace }}/.deps/install
+      
+      - name: Setup Nargo (only when folder is noir)
```
alxkzmn: I think we shouldn't restrict the version here to the same as the ProveKit is using. Please modify https://github.com/privacy-ethereum/csp-benchmarks/blob/CSP-Q3-2025/.github/actions/install-nargo/action.yml to accept the version as an optional argument, so that ProveKit and barretenberg can use the latest compatible versions independently.

guorong009: Done ed52313

barretenberg/sha256_prove.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Change to workspace root
+cd "$WORKSPACE_ROOT_PATH"
+
+#### Step 1: Witness generation ####
+WITNESS_FILE="sha256_${INPUT_SIZE}.gz"
+nargo execute --prover-name $TOML_PATH --package "sha256" $WITNESS_FILE
+
+#### Step 2: bb prove ####
+bb prove -b "$CIRCUIT_PATH" -w "$WORKSPACE_ROOT_PATH/target/$WITNESS_FILE" --write_vk -o "$WORKSPACE_ROOT_PATH/target/"
```
alxkzmn: I realized that we shouldn't write the VK each time inside the benchmark. In a real application this would probably be done only once before the app distribution. Moreover, it does not perform real proving when writing the VK as per description: 
> Write the verification key of a circuit. The circuit is
                         constructed using quickly generated but invalid witnesses
                         (which must be supplied in Barretenberg in order to expand
                         ACIR black box opcodes), and no proof is constructed.

The performance difference is also significant, I observed it to be ~25% slower when writing the VK.
We need to modify the benchmarks so that the VK is generated only once in the beginning. I think we can introduce an optional additional step of preparing the proof specifically for verification. E.g., `benchmark.sh` is looking for a file `prove_for_verification.sh`. If the file isn't present, we run the existing code that's in the script, but if the file is present, we run something like
```
hyperfine --runs "$RUNS" \
      --prepare $PROVE_FOR_VERIFICATION_SH \
      "STATE_JSON=$VERIFIER_JSON_FILE bash $VERIFY_SH" \
      --export-json "$SYSTEM_DIR/hyperfine_${TARGET}_${INPUT_SIZE}_verifier_metrics.json"
``` 
Alternatively, we may only override the proving part, not the whole `prepare` of a hyperfine call: look for `prove_for_verification.sh` file and then do
```
hyperfine --runs "$RUNS" \
      --prepare "UTILS_BIN=$UTILS_BIN INPUT_SIZE=$INPUT_SIZE STATE_JSON=$VERIFIER_JSON_FILE bash $PREPARE_SH && STATE_JSON=$VERIFIER_JSON_FILE bash $PROVE_FOR_VERIFICATION_SH > /dev/null 2>&1" \
      "STATE_JSON=$VERIFIER_JSON_FILE bash $VERIFY_SH" \
      --export-json "$SYSTEM_DIR/hyperfine_${TARGET}_${INPUT_SIZE}_verifier_metrics.json"
```

guorong009: Thanks for info!
Done 44ac0c5

alxkzmn (dismissed):
LGTM!

.github/actions/install-nargo/action.yml
```diff
@@ -1,12 +1,18 @@
 name: "Install Nargo"
 description: "Installs the noir-lang toolchain"
 
+inputs:
+  version:
+    description: Version of Nargo/Noir binaries.
+    required: false
+    default: ""
```
alxkzmn: I think it's better if we either require a version input or have a default hardcoded version, otherwise if it's not required and no default version is provided it's easy to overlook.

.github/workflows/sh_benchmarks_parallel.yml
```diff
@@ -99,12 +100,26 @@ jobs:
...
         uses: ./.github/actions/install-ligero
         with:
           ligero_path: ./ligetron/ligero-prover
           install_prefix: ${{ github.workspace }}/.deps/install
+      
+      - name: Setup Nargo (only when folder is barretenberg)
+        if: ${{ matrix.folder == 'barretenberg' }}
+        uses: ./.github/actions/install-nargo
+        with:
+          version: "1.0.0-beta.10"
```
alxkzmn: The latest for the end of September would be beta-13 https://github.com/noir-lang/noir/releases?q=beta-13&expanded=true

barretenberg/osx_local_setup.sh
```diff
@@ -0,0 +1,68 @@
...
+elif [ -n "${BASH_VERSION-}" ]; then
+    # We are in bash
+    # On macOS, .bash_profile is often used first, .bashrc sometimes sourced from it
+    [ -f "${HOME}/.bashrc" ] && source "${HOME}/.bashrc"
+    [ -f "${HOME}/.bash_profile" ] && source "${HOME}/.bash_profile"
+else
+    # Unknown shell: you can fallback or warn
+    echo "Warning: Unknown shell, cannot source rc file automatically" >&2
+fi
+bbup
```
alxkzmn: Barretenberg version compatible with Noir beta 11 is 0.87.0, so it should be 
```
bbup -v 0.87.0
```
How I found it: bbup allows to automatically install a compatible version by passing a Noir version argument (see here at the end https://github.com/AztecProtocol/aztec-packages/blob/master/barretenberg/bbup/README.md)

alxkzmn: (please also change to 0.87.0 in the github action)

.github/actions/install-nargo/action.yml
guorong009: Okay, I'll set the default version as `1.0.0-beta.9`.

.github/workflows/sh_benchmarks_parallel.yml
guorong009: I see. 
The issue is that `beta-13` version nargo compiler fails to compile the sha256 circuits in both `provekit` and `noir+barretenberg` benchmarks.

I've tried many versions. 
Found that `provekit` compiles with `beta-9`, and `noir+barretenberg` compiles with `beta-9` and `beta-10`.

barretenberg/osx_local_setup.sh
guorong009: Ok, I thought just calling the `bbup` solves all problem since it automatically installs compatible version of backend for installed nargo compiler.
If you think that way, I'll specify the version.

.github/workflows/sh_benchmarks_parallel.yml
alxkzmn: This is strange because the circuits are very different. I have solved that problem for ProveKit though with Noir beta 11: https://github.com/privacy-ethereum/csp-benchmarks/pull/90/files. I guess the problem is similar for pure Noir/bb, you may need to update the sha256 lib version

barretenberg/osx_local_setup.sh
alxkzmn: Not sure about that, maybe it does, but there may be backwards compatibility so we can accidentally install the latest backend (if it's backwards-compatible) but not the end-of-Q3 version.

guorong009: Done 4de99e7

.github/workflows/sh_benchmarks_parallel.yml
guorong009: Done 4de99e7

.github/actions/install-nargo/action.yml
guorong009: Done 4de99e7

.github/workflows/sh_benchmarks_parallel.yml
brech1: ~~Provekit _might_ be failing on latest `bb` due to being outdated! I selected a working old `rev` for the last lock update~~

Just saw updated refs

alxkzmn: So will it work with Noir beta 13 if you update the sha lib version?

guorong009: Yes, it works. Done 465fc43

benchmark.sh
```diff
@@ -76,10 +77,17 @@ for target in "${TARGETS[@]}"; do
       --export-json "$SYSTEM_DIR/hyperfine_${TARGET}_${INPUT_SIZE}_prover_metrics.json"
 
     step "[$TARGET] Verifier (size ${INPUT_SIZE}):"
-    hyperfine --runs "$RUNS" \
-      --prepare "UTILS_BIN=$UTILS_BIN INPUT_SIZE=$INPUT_SIZE STATE_JSON=$VERIFIER_JSON_FILE bash $PREPARE_SH && STATE_JSON=$VERIFIER_JSON_FILE bash $PROVE_SH > /dev/null 2>&1" \
-      "STATE_JSON=$VERIFIER_JSON_FILE bash $VERIFY_SH" \
-      --export-json "$SYSTEM_DIR/hyperfine_${TARGET}_${INPUT_SIZE}_verifier_metrics.json"
+    if [[ -x "$PROVE_FOR_VERIY_SH" ]]; then
+      hyperfine --runs "$RUNS" \
+        --prepare "UTILS_BIN=$UTILS_BIN INPUT_SIZE=$INPUT_SIZE STATE_JSON=$VERIFIER_JSON_FILE bash $PREPARE_SH && STATE_JSON=$VERIFIER_JSON_FILE bash $PROVE_FOR_VERIY_SH > /dev/null 2>&1" \
```
alxkzmn, line 82: This needs an update to the contribution guide

alxkzmn (dismissed):
LGTM!

benchmark.sh
guorong009: Done 529e76f

alxkzmn (dismissed):

.github/workflows/sh_benchmarks_parallel.yml
```diff
@@ -115,13 +115,13 @@ jobs:
         if: ${{ matrix.folder == 'barretenberg' }}
         uses: ./.github/actions/install-nargo
         with:
-          version: "1.0.0-beta.13"
+          version: "1.0.0-beta.11"
```
alxkzmn: The latest version by the end of September is beta.13 and I think it should work with bb 0.87.0, it probably didn't work for you before you've updated the sha256 lib but should work fine now.

guorong009: Done 4c26b32

alxkzmn (approved):

brech1 (approved):

PR #93, opened 2025-10-15
feat: implement `cairo-m` (kakarot zkVM) sha256 benchmark

### LINKED ISSUE #69: Add Kakarot SHA-256 benchmark

guorong009:
## Related issues
- Resolve #69 

## NOTE
- Add the `.cargo/config.toml` to `cairo-m` and `binius64` benchmarks, for custom rustflags


guorong009:
I meet the blocker for this PR.
The issue is that `Proof` struct cannot be cloned or copied. 
I raised the issue for this in `cairo-m` repo: https://github.com/kkrt-labs/cairo-m/issues/344

ClementWalter:
issues just solved and close, new patched release https://github.com/kkrt-labs/cairo-m/releases/tag/v0.1.0-alpha.1

guorong009:
Another blocker for PR: only one nightly version can build the code.
Raised the issue in the repo: https://github.com/kkrt-labs/cairo-m/issues/346
This error prevents the CI to be run.

alxkzmn:
@guorong009 if they can't update the toolchain, we exclude it from workspace and build and run separately.

.github/workflows/rust_benchmarks_parallel.yml
alxkzmn: I believe you need to add the cairo-m dir here, otherwise the CI will also try to bench it in the matrix later as the other Rust crates and will fail.

cairo-m/programs/sha256.cm
```diff
@@ -0,0 +1,123 @@
...
+    H[4] = H[4] + e;
+    H[5] = H[5] + f;
+    H[6] = H[6] + g;
+    H[7] = H[7] + h;
+    return ();
+}
+
+/// SHA-256 hash for 128-byte inputs (3 chunks after padding)
+/// Input: 48 u32 words representing pre-padded 32-byte message
+fn sha256_128(padded_message: [u32; 48]) -> [u32; 8] {
```
alxkzmn: Can we make this dynamic? Similarly to what we have in a few other circuits, we can take an array of the max possible input size (e.g., 2048) but only hash a part of it that's defined by an input size argument?

Cargo.toml
```diff
@@ -37,3 +40,7 @@ utils = { path = "utils" }
 ark-ff = { git = "https://github.com/a16z/arkworks-algebra", branch = "v0.5.0-optimize-mul-u64" }
 ark-ec = { git = "https://github.com/a16z/arkworks-algebra", branch = "v0.5.0-optimize-mul-u64" }
 ark-serialize = { git = "https://github.com/a16z/arkworks-algebra", branch = "v0.5.0-optimize-mul-u64" }
+
+# # patch for "cairo-m"
```
alxkzmn, line 45: Is this still needed?

cairo-m/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub fn prepare(input_size: usize) -> (RunnerOutput, Program) {
+    // Compile the program
+    let source_path = "programs/sha256.cm".to_string();
+    let source_text = fs::read_to_string(&source_path).expect("Failed to read sha256.cm");
+    let options = CompilerOptions {
+        verbose: false,
+        optimization_level: Default::default(),
+    };
+    let output =
```
alxkzmn, line 46: Currently, the `prepare` function compiles the guest program every time. This potentially affects the RAM measurement and also increases the total CI running time. This problem was solved in https://github.com/privacy-ethereum/csp-benchmarks/pull/94 - please take a similar approach, use the shared state closure to compile the guest and then only read the pre-compiled guest inside `prepare()`.

Cargo.toml
guorong009: No, I leave them since it would be needed when we add the `cairo-m` as workspace member in the future.
Fyi, I spent some hours because I didn't add this patch in workspace-level Cargo.toml.
I thought I'd not make such mistake again, when re-add the `cairo-m` as workspace member.

cairo-m/programs/sha256.cm
guorong009: Hmm, let me check the `cm` lang about that.

Cargo.toml
alxkzmn: I see, please add a TODO that explains that we need to re-enable it when we add cairo-m to the workspace

.github/workflows/rust_benchmarks_parallel.yml
guorong009: Done 32c9ec5

.github/workflows/rust_benchmarks_parallel.yml
```diff
@@ -229,3 +229,73 @@ jobs:
...
+          echo 'export CXX=/opt/homebrew/opt/llvm/bin/clang++' >> ~/.bashrc
+          echo 'export AR=/opt/homebrew/opt/llvm/bin/llvm-ar' >> ~/.bashrc
+          echo 'export RANLIB=/opt/homebrew/opt/llvm/bin/llvm-ranlib' >> ~/.bashrc
+
+          source ~/.bashrc
+
+      - name: Build "cairo-m" for memory report (temporary workaround)
+        run: |
+          cd cairo-m
+          RUSTFLAGS="-C link-arg=-fuse-ld=/opt/homebrew/bin/ld64.lld -C target-cpu=native" cargo build --release
```
alxkzmn: Is it possible to get rid of this extra step by providing the RUSTFLAGS into our benchmarking harness? The harness then would pass it to the binary compilation here:
https://github.com/privacy-ethereum/csp-benchmarks/blob/CSP-Q3-2025/utils/src/bench.rs#L175-L183

alxkzmn: So you could create one more harness signature variant that takes an extra argument, so that the other benchmarks are not affected.

guorong009: I think not.
When I try this command in my local machine, I use different `lld `path - `/usr/local/bin/ld64.lld`.
I believe there are some diffs between local environment and github CI.

alxkzmn: Have you tried `-fuse-ld=lld`? This is supposed to find the correct lld without specifying the absolute path. If that works, you can add the optional arg to the rust binary compilation.

Cargo.toml
guorong009: Done 3a2b403

cairo-m/programs/sha256.cm
guorong009: Done f64d6f2

.github/workflows/rust_benchmarks_parallel.yml
guorong009: Thanks for tip! It works! 👍 
I resolved the issue, but in different way.
According to Rust doc(https://rustwiki.org/en/cargo/reference/config.html?utm_source=chatgpt.com#configuration-format), we can set the config file for extra rustflags.
I add such config.toml file in the `cairo-m` dir.
Now, we can remove those custom rustflags.

Done 585b0c8

cairo-m/src/lib.rs
guorong009: I tried to follow the approach of shared state, and did some experiments.
Found that the intermediate state(compiled cairo program artifact) does not support proper serde.
It is a blocker similar to that `Proof` struct cannot be cloned or copied.
Maybe, I think we should leave the TODO note here. And move forward.

.github/workflows/rust_benchmarks_parallel.yml
alxkzmn: I see, but why does Binius64 need the flags now?

alxkzmn: Can we now remove the RAM measurement binary build from CI?

cairo-m/src/lib.rs
alxkzmn: Do you mean CompilerOutput::program? I see that it is serializable.

.github/workflows/rust_benchmarks_parallel.yml
guorong009: Ok, I'll remove the flags from `binius64`.

Regarding the `cairo-m`, we cannot remove the binary build from CI.
The reason is that it includes a temporary workaround(copy the binary to workspace `target` dir).
Once the `cairo-m` becomes a workspace member, we can remove the binary build from CI.

cairo-m/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+        .collect();
+    // padding adds extra 64 bytes to the input message bytes
+    let num_chunks = input_size / 64_usize + 1;
+    let runner_inputs = vec![
+        InputValue::List(input_values),
+        InputValue::Number(num_chunks as i64),
+    ];
+
+    // Run/Execute the program
+    let runner_output = run_cairo_program(
```
alxkzmn, line 79: Program execution is a part of proving process. Think about a real application: you wouldn't know the input until the moment you need to generate the proof. Therefore, you cannot precompute the execution trace in advance, it needs to be benchmarked inside the `prove()`, similar to what we do with witness generation for non-zkVM.

.github/workflows/rust_benchmarks_parallel.yml
alxkzmn: Got it!

cairo-m/src/lib.rs
guorong009: Yes, right. 
But, the inner fields are not implementing the serde.
Simply, when trying to serialize/deserialize from file, the serde of `CompilerOutput::program` fails.

cairo-m/src/lib.rs
guorong009: Done 9b120a6

cairo-m/src/lib.rs
alxkzmn: I see, let's leave it as is for now but I think we can coordinate with @ClementWalter to add full serialization support.

cairo-m/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    let output =
+        compile_cairo(source_text, source_path, options).expect("Failed to compile sha256.cm");
+    let compiled_program = (*output.program).clone();
+
+    (compiled_program, input_size)
+}
+
+pub fn prove(compiled_program: &Program, input_size: usize) -> Proof<Blake2sMerkleHasher> {
+    // Generate input using sha2_input
+    let (input_bytes, _digest) = generate_sha256_input(2048);
```
alxkzmn, line 51: Please build the input in `prepare()` so we don't benchmark it. We assume that the input is externally provided to the program, so we don't care about its generation time. Supplying `runner_inputs` into `prove()` would be enough, this way we only benchmark the trace/witness generation + proving.

guorong009: Done 343d8d1

cairo-m/src/lib.rs
guorong009: Yes, agree.
Created the issue in `cairo-m` repo: https://github.com/kkrt-labs/cairo-m/issues/348

Cargo.toml
brech1: I would create an `Add cairo-m to workspace` issue with this on it's description and cleanup these and other comments about it

alxkzmn (approved):
Looks good, let's proceed as discussed!

brech1 (approved):

PR #95, opened 2025-10-17
feat: implement `nexus` benchmarks

brech1:
# Description

Implements Nexus zkVM benchmarks for SHA256.

To be merged after:
  - https://github.com/eth-act/ere/pull/171

## Changes

  - Add Nexus zkVM benchmark implementation with guest program
  - Update ERE dependency from git rev to tagged release `v0.0.13`
  - Unify CI workflow handling for crates with custom toolchains (nexus, cairo-m)
  - Remove separate `bench_cairo_m` job for a unified excluded crates handling
  - Add LLVM/LLD installation separate GitHub action for cairo-m
  - Add separate target directories to CI cache for excluded crates

## Related Issues

- #73


nexus/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    // Convert ere_zkvm_interface::Proof (enum) to Vec<u8>
+    let proof_bytes = match proof {
+        ere_zkvm_interface::Proof::Compressed(bytes) => bytes,
+        _ => panic!("Nexus only supports Compressed proofs"),
+    };
+
+    ProofArtifacts {
+        proof_bytes,
+        public_values,
+        total_cycles: report.proving_time.as_secs(),
```
alxkzmn: Is this a workaround? Can't get the actual cycles? I think it would be better to report zero in this case.

alxkzmn: I realized my question may be incorrect, but I see that you're obtaining the cycles in a different way elsewhere, so this total_cycles in the struct is not necessary, isn't it?

nexus/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    prepared.compiled_size
+}
+
+/// Get the proof size from proof artifacts.
+pub fn proof_size(proof: &ProofArtifacts) -> usize {
+    proof.proof_bytes.len()
+}
+
+pub fn prepare_sha256(input_size: usize) -> PreparedSha256 {
+    let guest_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
alxkzmn: Currently, the prepare function compiles the guest program every time. This potentially affects the RAM measurement and also increases the total CI running time. This problem was solved in https://github.com/privacy-ethereum/csp-benchmarks/pull/94 - please take a similar approach, use the shared state closure to compile the guest and then only read the pre-compiled guest inside prepare().

nexus/src/lib.rs
brech1: Yeah sorry, I need to update this! Got fixed in https://github.com/eth-act/ere/pull/171

nexus/src/lib.rs
brech1: Got it, will update

brech1:
@alxkzmn Do you think the current approach of excluding nexus from the workspace and updating the CI makes sense? Unfortunately, nexus requires an older nightly version than the one that our current `toolchain` file pins. IMO this is the less invasive approach.

alxkzmn:
@brech1 i think it makes total sense and @guorong009 had to do the same for Cairo-M for the same reason (it needs the same nightly toolchain).

utils/src/lib.rs
```diff
@@ -23,7 +23,7 @@ pub fn write_json<T: Serialize>(data: &T, output_path: &str) {
 
 pub fn generate_sha256_input(input_size: usize) -> (Vec<u8>, Vec<u8>) {
     let mut message_bytes = vec![0u8; input_size];
-    let mut rng = rand::thread_rng();
+    let mut rng = StdRng::seed_from_u64(input_size as u64);
```
alxkzmn, line 26: Do you think we need to do this for all other benches as well?

alxkzmn (dismissed):
LGTM! Let's create an issue to track the results collection bug.

utils/src/lib.rs
brech1: Since these are benchmarks I think it's better to have seed `rngs`, but no strong opinions

brech1:
@alxkzmn I added a final commit to fix the collect, rest is unchanged, if it doesn't work I'll create the issue and fix on next PR

guorong009 (approved):

PR #107, opened 2025-10-22
Fail on result collection failure

alxkzmn:
- Result collection program will return error if any errors encountered during collection (will still write the valid results);
- Result collection workflow will fail if the collection program failed;
- Removed the unused manual trigger and wait loop (sometimes caused false failures);
- Made small changes in two benchmarks to trigger them and test the modified collection.


guorong009 (approved):

alxkzmn:
You can check out the runs in my fork (won't run here until merged):
- First run (after non-Rust benches finish) exits because Rust benches are still running: https://github.com/alxkzmn/csp-benchmarks/actions/runs/18735799127/job/53442034681#step:6:48
- Second run (after Rust benches finish) collects successfully: https://github.com/alxkzmn/csp-benchmarks/actions/runs/18737924252
- No wait loop necessary.

alxkzmn:
If result collection fails, it will look like this: https://github.com/alxkzmn/csp-benchmarks/actions/runs/18735666981/job/53441636451#step:11:80

brech1 (approved):

PR #116, opened 2025-10-28
Set full SHA256 input range; skip optional benchmarks

### LINKED ISSUE #85: Set SHA-256 benchmark input sizes array
Now the array contains only one value (128B: https://github.com/privacy-ethereum/csp-benchmarks/blob/CSP-Q3-2025/utils/src/metadata.rs). We need to set the array to our desired set of values (TBD, can be 128, 256, 512, 1024, 2048, 4096). Things to note:
- Circom benchmark is using pre-generated circuit files that are fixed for a particular size. We need to generate the additional circuits when setting the final input sizes array.

alxkzmn:
- SP1, OpenVM and Jolt benchmarks don't run by default - only on manual dispatch if the input is set;
- Benchmarks _on PR_ run with the "reduced" input set, e.g. 128B and 256B for SHA-256 (same as current);
- Benchmarks _on PR merge to CSP-Q3-2025 branch_ run with the full input set of 128 - 2048B

### DIFF
--- a/.github/workflows/dispatch_benchmarks.yml
+++ b/.github/workflows/dispatch_benchmarks.yml
@@ -11,6 +11,11 @@ on:
         required: false
         default: true
         type: boolean
+      include_optional:
+        description: "Include the mobile-unfriendly systems (openvm, jolt, sp1)"
+        required: false
+        default: false
+        type: boolean
 
 permissions:
   actions: write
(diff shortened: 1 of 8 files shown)


guorong009 (commented):
I think all good. Pls fix the CI errors and I'll approve the PR.

guorong009 (approved):

brech1 (approved):

PR #129, opened 2025-10-30
feat: add `BenchProperites` for `cairo-m`

### LINKED ISSUE #119: Report additional properties for Cairo-M
Add references to sources in the comments like this:
```rust
BenchProperties {
        proving_system: Some("Binius64".to_string()),
        field_curve: Some("GHASH binary field".to_string()), // https://www.binius.xyz/basics/binius64-vs-v0
        iop: Some("Binius64".to_string()),
        pcs: Some("Binius64".to_string()),
        arithm: Some("Binius64".to_string()),
        is_zk: Some(true),       // https://www.binius.xyz/basics/binius64-vs-v0
        security_bits: Some(96), // https://github.com/IrreducibleOSS/binius64/blob/main/verifier/verifier/src/verify.rs#L40
        is_pq: Some(true),       // hash-based PCS
        is_maintained: Some(true),
        is_audited: Some(AuditStatus::NotAudited),
        isa: None,
    },
```

guorong009:
## Related issues 
- Resolve #119

### DIFF
--- a/cairo-m/benches/sha256_bench.rs
+++ b/cairo-m/benches/sha256_bench.rs
@@ -2,14 +2,26 @@ use cairo_m::{prepare, prove, verify};
 use cairo_m_common::{InputValue, Program};
 use cairo_m_prover::{adapter::import_from_runner_output, public_data::PublicData};
 use cairo_m_runner::run_cairo_program;
-use utils::harness::ProvingSystem;
+use utils::harness::{AuditStatus, BenchProperties, ProvingSystem};
 
 utils::define_benchmark_harness!(
     BenchTarget::Sha256,
     ProvingSystem::CairoM,
     None,
     "sha256_mem_cairo_m",
-    utils::harness::BenchProperties::default(),
+    BenchProperties {
+        proving_system: Some("stwo/CSTARK/CircleSTARK".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        field_curve: Some("M31".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        iop: Some("circle IOPP".to_string()), // https://eprint.iacr.org/2024/278.pdf
+        pcs: Some("circle FRI".to_string()), // https://eprint.iacr.org/2024/278.pdf
+        arithm: Some("AIR".to_string()),
+        is_zk: Some(false),
+        security_bits: Some(96), // https://github.com/kkrt-labs/cairo-m/blob/main/crates/prover/src/prover_config.rs#L13-L20
+        is_pq: Some(true),       // hash-based PCS
+        is_maintained: Some(true), // https://github.com/kkrt-labs
+        is_audited: Some(AuditStatus::NotAudited), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#about
+        isa: None,
+    },
     |input_size| { prepare(input_size) },
     |_| 0,
     |(program, (entrypoint_name, runner_inputs))| {
--- a/cairo-m/benches/sha256_bench.rs
+++ b/cairo-m/benches/sha256_bench.rs
@@ -10,10 +10,10 @@ utils::define_benchmark_harness!(
     None,
     "sha256_mem_cairo_m",
     BenchProperties {
-        proving_system: Some("stwo/CSTARK/CircleSTARK".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        proving_system: Some("Circle STARK".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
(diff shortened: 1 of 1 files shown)


cairo-m/benches/sha256_bench.rs
```diff
@@ -2,14 +2,26 @@ use cairo_m::{prepare, prove, verify};
...
+    BenchProperties {
+        proving_system: Some("stwo/CSTARK/CircleSTARK".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        field_curve: Some("M31".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        iop: Some("circle IOPP".to_string()), // https://eprint.iacr.org/2024/278.pdf
+        pcs: Some("circle FRI".to_string()), // https://eprint.iacr.org/2024/278.pdf
+        arithm: Some("AIR".to_string()),
+        is_zk: Some(false),
+        security_bits: Some(96), // https://github.com/kkrt-labs/cairo-m/blob/main/crates/prover/src/prover_config.rs#L13-L20
+        is_pq: Some(true),       // hash-based PCS
+        is_maintained: Some(true), // https://github.com/kkrt-labs
```
brech1, line 21: This is a tricky one, not for this vm in particular but some might not be as active but still maintained. BTW, shouldn't this be `https://github.com/kkrt-labs/cairo-m` instead?

brech1 (dismissed):

cairo-m/benches/sha256_bench.rs
guorong009: I think the maintenance status is different from activeness. It can be active or not.

Regarding the URL, I thought we should put the url of team behind the maintenance - not the repo itself.

cairo-m/benches/sha256_bench.rs
```diff
@@ -2,14 +2,26 @@ use cairo_m::{prepare, prove, verify};
...
 utils::define_benchmark_harness!(
     BenchTarget::Sha256,
     ProvingSystem::CairoM,
     None,
     "sha256_mem_cairo_m",
-    utils::harness::BenchProperties::default(),
+    BenchProperties {
+        proving_system: Some("stwo/CSTARK/CircleSTARK".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        field_curve: Some("M31".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
+        iop: Some("circle IOPP".to_string()), // https://eprint.iacr.org/2024/278.pdf
```
alxkzmn: Let's specify both IOP and PCS as "Circle-FRI" (with the capital C) because FRI is also an integral part of the IOP here.

cairo-m/benches/sha256_bench.rs
alxkzmn: @brech1 I think for now it's fine to have it coarse like this, it's difficult to come up with a metric for a "degree of activity" so we can simply say maintained or not.

cairo-m/benches/sha256_bench.rs
```diff
@@ -2,14 +2,26 @@ use cairo_m::{prepare, prove, verify};
...
+use utils::harness::{AuditStatus, BenchProperties, ProvingSystem};
 
 utils::define_benchmark_harness!(
     BenchTarget::Sha256,
     ProvingSystem::CairoM,
     None,
     "sha256_mem_cairo_m",
-    utils::harness::BenchProperties::default(),
+    BenchProperties {
+        proving_system: Some("stwo/CSTARK/CircleSTARK".to_string()), // https://github.com/kkrt-labs/cairo-m/?tab=readme-ov-file#welcome-to-cairo-m
```
alxkzmn: This one is not a big deal, but I think we can make some rules for ourselves that simplify the specification here, and make the labels shorter so that they better fit into the table in EthProofs. We can default to the proving system name from the original paper, and use the implementations name only if the original paper does not introduce a specific name for the proving system. In this case, the paper introduces "Circle STARK". CSTARK abbreviation is arguably not widely used/recognized. STWO is a name of an "intermediate" implementation codebase of Circle STARK that the resulting Cairo-M is using, but Cairo-M itself has lots of code that affects the resulting system properties, so Cairo-M/STWO are the implementation details. Therefore, we can specify the underlying proving system as "Circle STARK".

guorong009: Done [ca62b9f](https://github.com/privacy-ethereum/csp-benchmarks/pull/129/commits/ca62b9f4b825e28c2fa2971ae27457e42a435c5f)

cairo-m/benches/sha256_bench.rs
guorong009: Done [ca62b9f](https://github.com/privacy-ethereum/csp-benchmarks/pull/129/commits/ca62b9f4b825e28c2fa2971ae27457e42a435c5f)

brech1 (approved):

brech1:
I think `BenchProperties` `String`s could be updated to `&'static str` since these are constant properties

alxkzmn:
@brech1 agreed, let's do it in the next PR.

PR #138, opened 2025-10-31
Implement ECDSA benchmark for ProveKit

### LINKED ISSUE #137: Implement ProveKit ECDSA benchmark

alxkzmn:
- Updated the circuit from https://github.com/worldfnd/provekit/tree/main/noir-examples/p256_bigcurve for compatibility with the our Q3 Noir version
- Added the ECDSA input generator that is mean to be reused by all benchmarks
- Removed the unused cfg.fixed_input_size in harness


guorong009 (commented):
I think we don't need to push [`Prover.toml`](provekit/circuits/ecdsa/p256_bigcurve/Prover.toml) file. 
Can we ignore that file in `.gitignore` inside the `provekit` dir?

alxkzmn:
@guorong009 having it is still good for the cases when you need to debug the circuit without regenerating the input, and it will work out of the box for anyone who checks out the code. Also, writing out the generated values during the benchmark only requires replacing the values, not creating the file from scratch.

guorong009 (approved):

brech1 (approved):

PR #146, opened 2025-11-05
feat: implement ECDSA benchmark in  "barretenberg"

### LINKED ISSUE #126: Report additional properties for Barretenberg
An easy way to pass all properties at once is via a JSON file:
https://github.com/privacy-ethereum/csp-benchmarks/blob/f9c701a3cd5cbd7ce15953c702505830f3772bfe/utils/src/bin/format_hyperfine.rs#L76-L78
Recommended to resolve together with #127

### LINKED ISSUE #131: Implement Noir/BB ECDSA benchmark
Circuit: https://github.com/noir-lang/noir/blob/master/noir_stdlib/src/ecdsa_secp256k1.nr#L1-L23

guorong009:
## Related issues
- Resolve #131 
- Resolve #126 

## Changes
- add noir circuit of ECDSA benchmark 
- add shell scripts for ECDSA benchmark
- improve existing shell scripts for ECDSA benchmark (`benchmark.sh`)
- improve `utils::main` binary for ECDSA benchmark
- add `bench_props.json` for `barretenberg`


utils/src/lib.rs
```diff
@@ -49,6 +49,13 @@ pub fn generate_ecdsa_input() -> (Vec<u8>, (Vec<u8>, Vec<u8>), Vec<u8>) {
     let signature: Signature = signing_key
         .sign_prehash(&digest)
         .expect("Failed to sign prehashed digest");
+
+    // Normalize "s" of the signature because it is not normalized by default.
+    // More importantly, the "noir::std::ecdsa_secp256r1::verify_signature" expects "s" to be normalized.
+    let signature = signature
```
alxkzmn, line 57: Have you tested it for ProveKit?

benchmark.sh
alxkzmn: If we modify the utils binary to accept the bench_target argument and return a single size of "32" for ECDSA, we can use the existing code with minimal modifications. Rust harness is actually doing this trick - all Rust ECDSA benches are running with sizes = [32] and are structurally equivalent to sha256 benches.

alxkzmn (commented):
If we modify the utils CLI to return a single "32" size for ECDSA (similar to Rust harness), we won't need extensive modifications of the `benchmarks.sh`. This would be beneficial because handling ECDSA as a separate case in `benchmarks.sh` leads to code substantial duplication.

benchmark.sh
guorong009: Done 3b490cc

utils/src/lib.rs
guorong009: Yes, I've tested it in provekit. It's working.

alxkzmn (approved):
LGTM! Great job!

brech1 (approved):

PR #148, opened 2025-11-06
Report number of constraints for non-Rust systems

### LINKED ISSUE #143: Report number of constraints for non-Rust systems
Consider the following approaches:
- If it is possible compile the circuit during the benchmark, or to get the number of constraints from the pre-compiled circuit - introduce a new shell script in each system's directory that will do that and will be called by `benchmark.sh`
- If it's not possible to easily access the number of constraints from shell script, report them via CLI when calling the `format_hyperfine`. In this case they may be provided in an additional JSON file (e.g., `num_constraints.json` in each system's dir, containing a mapping by bench type -> circuit size -> num constraints).

alxkzmn:
Changes:
- reporting the number of circuit constraints for non-Rust systems by passing the manually measured values via a JSON file. The reasoning is that for non-Rust systems we typically don't have an easy access to these numbers - they're often printed in the console as a by-product of the circuit compilation; in case of Noir you even need to run a separate command, `noir-profiler`, to get them. It's still possible to parse all those command outputs somewhere in `..._prepare.sh` but that would be an overkill IMHO. The middle ground is to keep the output logs of invoking those commands. Although up for discussion.

Fix:
- tailor Noir/BB & ProveKit circuits to the exact input size. Our initial implementation was in fact compiling a circuit of the same size (maximum, 2048) for all input sizes which defeated the purpose of varying the size.


alxkzmn:
Noir/BB bench will go through after merging https://github.com/privacy-ethereum/csp-benchmarks/pull/146 (currently missing the `bench_props.json`)

brech1 (dismissed):
I agree that ideally we should generate these dynamically from prover output instead of dealing with pre-build static files, and understand that it would need custom parse files for each system's logs, probably overkill right now. Current approach works! But if you also believe dynamic generation is a good path forward we could track this in an issue.

brech1 (approved):

PR #152, opened 2025-11-10
feat: add optimization flag in `binius64` compile

### LINKED ISSUE #115: Enable native optimizations for Binius64
Binius64 has some native optimizations and would potentially demonstrate better performance if compiled with 
```bash
export RUSTFLAGS="-C target-cpu=native"
```
Currently this fails because of the following error:
https://github.com/privacy-ethereum/csp-benchmarks/actions/runs/18672101870/job/53237027330#step:10:429
(the `ring` crate fails to compile in an unrelated example that we don't need but still depend on if depending on the whole examples crate). 
This is a blocker for compiling with optimizations.

guorong009:
## Related issues
- Resolve #115

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,17 +100,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
-name = "aes"
-version = "0.8.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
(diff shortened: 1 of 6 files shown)


alxkzmn (changes requested):
All good, let's just remove the word "example" from all the struct/variable/file names - it is an example in the context of the Binius64 repo but it's a legit standalone SHA-256 implementation in our context so we can simply call it a SHA-256 circuit or something.

guorong009:
> All good, let's just remove the word "example" from all the struct/variable/file names - it is an example in the context of the Binius64 repo but it's a legit standalone SHA-256 implementation in our context so we can simply call it a SHA-256 circuit or something.

Done 0a4392e

alxkzmn (approved):
LGTM!

brech1 (approved):

PR #153, opened 2025-11-10
Make reported properties non-optional

### LINKED ISSUE #118: Make reported properties non-optional
https://github.com/privacy-ethereum/csp-benchmarks/blob/f9c701a3cd5cbd7ce15953c702505830f3772bfe/utils/src/harness.rs#L113-L132
BenchProperties struct should only have the optional fields according to the table here: https://github.com/privacy-ethereum/csp-benchmarks/issues/66
This issues should be resolved after all the child issues are resolved.

alxkzmn:


alxkzmn:
As there were some changes to the non-Rust harness, I've triggered the Ligetron bench via a small change.

brech1 (approved):

PR #176, opened 2025-12-01
Add Q3 report final data

alxkzmn:
Collected in https://github.com/privacy-ethereum/csp-benchmarks/actions/runs/19450978214


brech1:
I would rather not include the file directly in the repository since it's an auto-generated, non editable file, and usually won't be the result of the current state of the project.

My suggestion to deal with this is that we create a new simple manual workflow, similar to the `supabase_upload` one, that receives the AWS run ID and a label (e.g. Q3-25) and generates a release and appends this file as a download attachment (like some projects include their binaries).

This would tie the file release to the given commit (project status) and job run details.

Let me know your thoughts

alxkzmn:
We can attach the zipped file if we create a release here on Github but this can be done in parallel with committing the original data.

brech1 (approved):

PR #190, opened 2026-01-08
feat(utils): add Poseidon input generators and bench target

brech1:
# Description

Adds shared utilities for Poseidon benchmarks


alxkzmn (approved):
LGTM!

PR #192, opened 2026-01-08
feat(plonky2): add Poseidon benchmarks

brech1:
# Description

Adds Poseidon benchmarks for Plonky2

### DIFF
--- a/plonky2/benches/poseidon.rs
+++ b/plonky2/benches/poseidon.rs
@@ -1,17 +1,16 @@
-use plonky2_sha256::bench::{prove, sha256_prepare, verify};
-
-use plonky2::{plonk::config::PoseidonGoldilocksConfig, util::serialization::Write};
-use plonky2_u32::gates::arithmetic_u32::{U32GateSerializer, U32GeneratorSerializer};
+use plonky2::plonk::config::PoseidonGoldilocksConfig;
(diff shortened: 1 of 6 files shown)


plonky2/benches/poseidon.rs
```diff
@@ -0,0 +1,54 @@
...
+
+const D: usize = 2;
+type C = PoseidonGoldilocksConfig;
+
+utils::define_benchmark_harness!(
+    BenchTarget::Poseidon,
+    ProvingSystem::Plonky2,
+    None,
+    "poseidon_mem_plonky2",
+    BenchProperties::new(
```
alxkzmn: The properties could be defined as a constant only once and reused across all benchmarks, similar to this:
https://github.com/privacy-ethereum/csp-benchmarks/blob/6ca21a60aba1694177f0e88cdd6281d93a7295ea/binius64/src/lib.rs#L27-L39
...
https://github.com/privacy-ethereum/csp-benchmarks/blob/6ca21a60aba1694177f0e88cdd6281d93a7295ea/binius64/benches/sha256_bench.rs#L10-L17

alxkzmn (changes requested):
All good, just need to extract the Plonky2 properties as a constant and reuse across all benches.

alxkzmn (approved):

PR #193, opened 2026-01-08
feat(barretenberg): add Poseidon benchmarks

brech1:
# Description

Adds Poseidon benchmarks for Barretenberg


barretenberg/verify.sh
alxkzmn, line 1: We could reuse the whole verify script as is because it's identical across all BB benches. You could try to do it in this PR if you don't mind. Same goes for prove, prove_for_verify and verify, given that we pass the bench name as an argument to substitute the filenames.

alxkzmn (changes requested):
Let's maximize the reuse of prove, prove_for_verify and verify.

benchmark.sh
```diff
@@ -107,17 +106,10 @@ for target in "${TARGETS[@]}"; do
...
-        "STATE_JSON=$VERIFIER_JSON_FILE bash $VERIFY_SH" \
-        --export-json "$SYSTEM_DIR/hyperfine_${TARGET}_${INPUT_SIZE}_verifier_metrics.json"
-    else
-      hyperfine --runs "$RUNS" \
-        --prepare "UTILS_BIN=$UTILS_BIN INPUT_SIZE=$INPUT_SIZE STATE_JSON=$VERIFIER_JSON_FILE bash $PREPARE_SH && STATE_JSON=$VERIFIER_JSON_FILE bash $PROVE_SH > /dev/null 2>&1" \
-        "STATE_JSON=$VERIFIER_JSON_FILE bash $VERIFY_SH" \
-        --export-json "$SYSTEM_DIR/hyperfine_${TARGET}_${INPUT_SIZE}_verifier_metrics.json"
-    fi
+    hyperfine --runs "$RUNS" \
+      --prepare "UTILS_BIN=$UTILS_BIN INPUT_SIZE=$INPUT_SIZE STATE_JSON=$VERIFIER_JSON_FILE bash $PREPARE_SH && WRITE_VK=1 STATE_JSON=$VERIFIER_JSON_FILE bash $PROVE_SH > /dev/null 2>&1" \
```
alxkzmn: WRITE_VK is specific to Barretenberg, it shouldn't be in the top-level benchmarking script (e.g., it wouldn't make sense for Ligetron)

alxkzmn (approved):
LGTM!

PR #200, opened 2026-01-14
feat(circom): add Keccak benchmarks

brech1:
# Description

Adds Circom Keccak benchmarks.


utils/src/metadata.rs
```diff
@@ -17,3 +17,13 @@ pub fn selected_poseidon_inputs() -> Vec<usize> {
         _ => POSEIDON_INPUTS_FULL.to_vec(),
     }
 }
+
+const KECCAK_INPUTS_REDUCED: [usize; 2] = [128, 256];
```
alxkzmn: TBH I think it's ok to have the same inputs for sha and keccak (and any other "pre-ZK" hashes, if we add them).

.gitmodules
```diff
@@ -1,6 +1,9 @@
...
-path = ligetron/ligero-prover
-url = https://github.com/ligeroinc/ligero-prover.git
+	path = ligetron/ligero-prover
+	url = https://github.com/ligeroinc/ligero-prover.git
 [submodule "circom/circomlib"]
 	path = circom/circomlib
 	url = https://github.com/iden3/circomlib.git
+[submodule "circom/hash-circuits"]
+	path = circom/hash-circuits
+	url = https://github.com/bkomuves/hash-circuits
```
alxkzmn, line 9: Have you checked https://github.com/vocdoni/keccak256-circom? Which one has better performance?

alxkzmn: I also see in the repo that you have used that there is a bit version besides the byte version - sane question here, how does the performance compare?

brech1: I tried that one first but it stops working with inputs larger than 136 bytes (single input block). Seems like a leaner implementation aimed to work with Ethereum data mostly. The one I'm using now is the same thing but with a block loop for larger inputs.

The bits version is more efficient since it doesn't have the byte to bit overhead, like the wrapper we're using for SHA. For a fair comparison, both should use bytes, or I should remove the SHA wrapper and use bits here as well. But I guess it's more real-world to use bytes?

alxkzmn: I see, bytes is more general-purpose because we can't assume that a private message could be fed into the "parent" circuit (whose part the keccak circuit is) as bits and wouldn't be represented as bytes in any other part of the parent circuit. Simply speaking, there's a high chance that the conversion would happen anyways.

alxkzmn (approved):
LGTM!

PR #212, opened 2026-01-21
feat(risc0): add Keccak benchmarks

brech1:
# Description

Adds risc0 Keccak benchmarks

## Related Issues

- #209


alxkzmn:
@brech1 is this using a precompile? I see in the docs and the example that they say they're patching the `tiny-keccak` but not the `sha3`: https://dev.risczero.com/api/zkvm/precompiles#patched-crates, https://github.com/risc0/tiny-keccak/releases
The example: https://github.com/risc0/risc0/blob/main/examples/keccak/methods/guest/Cargo.toml

brech1:
> @brech1 is this using a precompile? I see in the docs and the example that they say they're patching the `tiny-keccak` but not the `sha3`: https://dev.risczero.com/api/zkvm/precompiles#patched-crates, https://github.com/risc0/tiny-keccak/releases The example: https://github.com/risc0/risc0/blob/main/examples/keccak/methods/guest/Cargo.toml

Oh, it's not, since this is code from their benchmarks didn't think there could be one. I'll update

utils/src/zkvm/sha256.rs
alxkzmn, line 1: Nice!

alxkzmn (approved):
LGTM!

Issue #144, opened 2025-11-04
Implement variable input `jolt` guests


brech1:
Ere added support for passing inputs in the `jolt` implementation:

- https://github.com/eth-act/ere/pull/181

The input size is limited to `4096` bytes.

This allows us to remove all the specific input size guests and update the programs to receive arguments.

alxkzmn:
@brech1 will this be closed by https://github.com/privacy-ethereum/csp-benchmarks/pull/217?

PR #217, opened 2026-01-29
chore: update ere and related zkvm versions

### LINKED ISSUE #178: Update Jolt version
- Latest Jolt version should have significantly improved performance: https://a16zcrypto.com/posts/article/64-bit-proving-jolt/
- After updating it we need to reassess its feasibility for CSP and potentially remove the exclusion https://github.com/privacy-ethereum/csp-benchmarks/blob/4de0cecbae97286e4390ef621f367b3abd7b8749/.github/workflows/rust_benchmarks_parallel.yml#L114

brech1:
# Description

Updates `ere` dependency and upgrades zkVM dependencies to latest versions.

## Changes

- `ere` updated from v0.0.13 to commit `aaf2efcc`
- `rust-toolchain` updated from nightly-2025-08-18 to nightly-2026-01-29
- `risc0` guests updated from 3.0.3 to 3.0.4
- `nexus` updated from v0.3.4 to v0.3.6, added `zmij` dependency
- Jolt patches switched from `v0.5.0-optimize-mul-u64` to `dev/twist-shout` branch
- `openvm` added `compiler` and `zkvm` features

---
Resolves #178

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2025-08-18-aarch64-apple-darwin
+          toolchain: nightly-2026-01-29
           override: true
           profile: default
           default: true
--- a/.github/workflows/rust_benchmarks_parallel.yml
+++ b/.github/workflows/rust_benchmarks_parallel.yml
@@ -153,7 +153,7 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2025-08-18-aarch64-apple-darwin
+          toolchain: nightly-2026-01-29
           override: true
           profile: default
           default: true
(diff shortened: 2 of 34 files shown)


alxkzmn:
@brech1 Looks good! Could you please check Jolt's performance locally against some "median" proving system? If the performance is as good as in their mobile demo (meaning if it's feasible for us to run it), we also need to remove Jolt from the exceptions and always run the benchmark by default.

brech1:
> @brech1 Looks good! Could you please check Jolt's performance locally against some "median" proving system? If the performance is as good as in their mobile demo (meaning if it's feasible for us to run it), we also need to remove Jolt from the exceptions and always run the benchmark by default.

I've been working in a fix for some current issues that jolt is giving with this PR update, but I believe I could include this with a proper resolve of #144 and

- #223 
- #222 

In a PR of its own to have a more informed decision. WDYT?

alxkzmn:
Sure, let's split it.

alxkzmn (approved):
LGTM!

PR #220, opened 2026-01-31
Add Spartan2 SHA-256 benchmark

### LINKED ISSUE #134: Implement Spartan2 benchmark
Spartan-ECDSA repo: https://github.com/personaelabs/spartan-ecdsa. Requires installing a custom version of Circom so will require changes to CI.

alxkzmn:
Closes #134


.github/actions/install-risc0/action.yml
```diff
@@ -1,10 +1,18 @@
...
+    description: "GitHub token for API authentication"
+    required: false
+    default: ""
+
 runs:
   using: "composite"
   steps:
     - shell: bash
+      env:
+        GITHUB_TOKEN: ${{ inputs.github-token }}
```
brech1: I think this might not be needed, the workflow already sets `GITHUB_TOKEN` at the top level (line 35 in `rust_benchmarks_parallel.yml`) so it should be available here, and it's currently working on other PRs

spartan2/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "spartan2-bench"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+spartan2 = { git = "https://github.com/microsoft/Spartan2.git", default-features = false }
```
brech1: We should pin this to a specific rev or tag

.github/workflows/rust_benchmarks_parallel.yml
```diff
@@ -174,6 +174,8 @@ jobs:
 
       - name: Install RISC0 toolchain
         uses: ./.github/actions/install-risc0
+        with:
+          github-token: ${{ secrets.GH_PAT }}
```
brech1: This can be removed as well

brech1 (dismissed):
LGTM, but I would remove the arg from the risc0 job since it's not expecting it

PR #224, opened 2026-02-04
feat(jolt): fix and add guests

### LINKED ISSUE #144: Implement variable input `jolt` guests

### LINKED ISSUE #222: Implement Jolt ECDSA benchmark
https://github.com/a16z/jolt/tree/main/examples/secp256k1-ecdsa-verify

### LINKED ISSUE #223: Implement Jolt Keccak-256 benchmark
https://github.com/a16z/jolt/blob/main/examples/sha3-ex/guest/src/lib.rs

brech1:
# Description

Upgrades Jolt benchmarks to use dynamic input sizes and adds Keccak-256 and ECDSA secp256k1 support. 

Removes Jolt from optional workflow (checking). 

Fixes SHA256 guest compiler.

Should be merged (and updated) after:

- https://github.com/eth-act/ere/pull/286

## Related Issues

- Resolves #144 
- Resolves #222  
- Resolves #223


alxkzmn:
@brech1 Jolt team reached out to me and pointed out that they have more inlines (including sha256 and keccak, see here: https://github.com/a16z/jolt/tree/main/jolt-inlines). Could you please include their usage in this PR?

jolt/src/lib.rs
```diff
@@ -1,25 +1,122 @@
...
+    verify_ecdsa, verify_keccak, verify_sha256,
 };
 
+pub fn jolt_bench_properties() -> BenchProperties {
+    BenchProperties::new(
+        "Lasso+Binius",
+        "Binary tower",
+        "sumcheck",
+        None,
+        "R1CS",
```
alxkzmn:
```suggestion
        "Jolt",
        "Bn254",
        "Twist & Shout",
        "Dory",
        "Jolt", 
```
The system name (the first one) is the "commercial" name that we call it => Jolt (otherwise we wouldn't find "Jolt" in the results table on EthProofs).
Curve and PCS - as instantiated in https://github.com/eth-act/ere/blob/79fb48d6f26e2a9abefb534428f27901810f1206/crates/zkvm/jolt/src/zkvm/sdk.rs#L10-L14 (the default import values from the Jolt SDK)
Lasso was replaced by Twist & Shout: https://jolt.a16zcrypto.com/how/architecture/architecture.html#jolts-five-components
The arithmetization is tricky, it's kind of custom here (also, R1CS cannot support lookups so it can't be just R1CS), so it's easier to call it "Jolt" too. The docs say R1CS is only used to handle the "fetch" part of the fetch-decode-execute loop: https://jolt.a16zcrypto.com/how/architecture/architecture.html#r1cs-constraints

jolt/src/lib.rs
```diff
@@ -1,25 +1,122 @@
...
+        true,
```
alxkzmn:
```suggestion
        false,
```
Not-yet-ZK, see: https://github.com/a16z/jolt/pull/1205

jolt/src/lib.rs
```diff
@@ -1,25 +1,122 @@
...
+        true,
+        128,
+        true,
```
alxkzmn:
```suggestion
        false,
```
Dory is not PQ-sound.

brech1:
Converting this to draft since it has to be merged after: 

- https://github.com/eth-act/ere/pull/290

alxkzmn (approved):
LGTM!

PR #229, opened 2026-02-09
feat(ligetron): add ecdsa benchmarks

jdodinh:


jdodinh:
@aesilevich added features that enable ECDSA verification with ligetron, and I am openingthis PR to integrate this implementation with the benchmarking framework.

This also updates the ligero-prover submodule to v1.4.0 and adjusts existing scripts to the new changes.

Available to answer any necessary questions.

CC @mvenkita

ligetron/osx_local_setup.sh
```diff
@@ -24,6 +24,10 @@ if [[ "$(uname -s)" != "Darwin" ]]; then
   echo "This script is for macOS (Darwin) only."; exit 1
 fi
 
+# Detect number of CPU cores for parallel builds (leave 2 cores free for system)
+NPROC=$(sysctl -n hw.ncpu)
```
alxkzmn: Is this necessary? cmake should be able to determine the necessary number of jobs automatically.

ligetron/osx_local_setup.sh
```diff
@@ -146,18 +154,16 @@ ok "emsdk ready (emcmake available)"
 # Ligetron submodule
 # -----------------------
 LIGETRON_DIR="${SCRIPT_DIR}/ligero-prover"
-if [[ ! -d "${LIGETRON_DIR}" ]]; then
```
alxkzmn, line 124: The guard was there to prevent the unnecessary update in case the directory is present, please don't remove.

alxkzmn:
@jdodinh the build is failing: https://github.com/privacy-ethereum/csp-benchmarks/actions/runs/21896590194/job/63215678065?pr=229

ligetron/osx_local_setup.sh
jdodinh: I recall having problems with installing Dawn with -j, and was recommended to specify 2 fewer cores than my machine, which made it work, so I got in the habit of doing that.

However, now that I think about it, the problem was on Ubuntu, so yes, I will remove it.

ligetron/osx_local_setup.sh
jdodinh: I will put the guard back, but will add an additional check to make sure the right version of Ligetron is there.

ligetron/osx_local_setup.sh
jdodinh: Brought back parallelism in 47cc5904ba46b9bab40a5af714e84b312974e315.

ligetron/osx_local_setup.sh
jdodinh: Done in 59e7c3067664bc584ca6f427f2009461e92b27c4.

jdodinh:
Hi @alxkzmn,

Workflow failed because it had old paths to the SDK. I updated these paths in 1e8f323bdd23ca2775de21956e97aee6977fd94d. 

In 263033dc15da3dcd2d828808efeed9bbe1a57f62 I updated the Wabt and Dawn versions that need to be used. I was experiencing problems with Wabt v1.0.39, so I enforced the use of v1.0.38 in the meantime.

Hopefully everything should work now :)

alxkzmn:
@jdodinh please take a look at https://github.com/privacy-ethereum/csp-benchmarks/pull/232 - we'd like to ask you to do a similar thing in your PR - re-use the common prepare, prove and verify scripts.

jdodinh:
Sounds good, I will do that.

jdodinh:
Hi @alxkzmn, I made the changes, and I also did a test run on a ligero fork of this repo. Everything passed: 

https://github.com/ligeroinc/csp-benchmarks/actions/runs/21998059013/job/63563287155?pr=1

brech1 (changes requested):
I think the PR is not correctly scoped, introducing many changes that might be consequence of submodule bump, and not adapting correctly to the changes in:

- #232

Is it possible to only introduce the necessary to run ECDSA benchmarks with the current setup? We can deal with the bump in a following PR

ligetron/measure.sh
```diff
@@ -17,17 +17,23 @@ TARGET_NAME="${BASENAME%%_[0-9]*}"
...
 
 # Preprocessing artifacts: WASM used by the prover
-WASM_PATH="${SCRIPT_DIR}/ligero-prover/sdk/build/examples/${TARGET_NAME}.wasm"
+# Map target name to WASM filename. Most targets use ${TARGET_NAME}.wasm, but some
+# (like ECDSA) have specialized variants with different names in the SDK.
+case "$TARGET_NAME" in
+  ecdsa) WASM_NAME="ecdsa_p256_verify_prehashed" ;;
+  *)     WASM_NAME="$TARGET_NAME" ;;
+esac
+WASM_PATH="${SCRIPT_DIR}/ligero-prover/sdk/cpp/build/examples/${WASM_NAME}.wasm"
```
brech1: This seems like an unnecessary case? not properly using the argument

.github/actions/install-ligero/action.yml
brech1, line 1: Changes on the workflow shouldn't be necessary

jdodinh:
@brech1 

Happy to move the version bump to a different PR.

Regarding the CI changes: the version bump requires them because the SDK paths changed. They also improve caching for Dawn and WABT and pin them to recommended versions. Is there anything specific you're concerned about with these changes? Happy to discuss.

For the ECDSA integration, I'll adapt once the version bump is merged.

jdodinh:
@brech1 
Opened #236.

.github/actions/install-ligero/action.yml
```diff
@@ -114,6 +119,29 @@ runs:
...
+        git submodule update --init
+        mkdir -p build && cd build
+        cmake -G Ninja \
+              -DCMAKE_CXX_COMPILER=clang++ \
+              -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+              -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+              -DCMAKE_INSTALL_PREFIX="${{ inputs.install_prefix }}" ..
+        cmake --build . --parallel
+        cmake --install .
+
```
alxkzmn: See https://github.com/privacy-ethereum/csp-benchmarks/pull/236#issuecomment-3900402362

ligetron/measure.sh
alxkzmn: I think this is because the sha circuit is "sha256.wasm" so it maps nicely, whereas for ecdsa it needs adjustment, is that correct @jdodinh ?

alxkzmn:
@jdodinh I see that Ligero has introduced the ECDSA circuit in a recent SDK version, does it mean we first need to merge #236 before proceeding here?

ligetron/measure.sh
jdodinh: @alxkzmn Yes precisely, while the prepare script allows to specify the specific wasm path, the `measure.sh` script defaults to `${TARGET_NAME}.wasm`, hence this workaround. Currently looking into making this cleaner.

jdodinh:
@alxkzmn @brech1 Just added the benchmarks for ECDSA. I adjusted `measure.sh` to retrieve the WASM program path from `STATE_JSON`, and if it's not present, it defaults to using `TARGET_NAME` to build the path (`measure.sh:30-31`):
```
WASM_PATH=$(jq -r '.program // empty' "$STATE_JSON")
WASM_PATH="${WASM_PATH:-${SCRIPT_DIR}/ligero-prover/sdk/cpp/build/examples/${TARGET_NAME}.wasm}"
```
With this setup, I actually think that the second line can be removed as the program entry in the JSON will always be present, but feel free to object.

ligetron/ecdsa_prepare.sh
alxkzmn, line 1: Should be using the `common_prepare.sh` in a similar way to https://github.com/privacy-ethereum/csp-benchmarks/blob/main/ligetron/sha256_prepare.sh

jdodinh: `common_prepare.sh` is built around a hashing pattern that ECDSA doesn't fit. I don't think that refactoring it to accommodate ECDSA falls into the scope of this PR. Would you still like me to proceed? Otherwise, happy to create an issue for it.

alxkzmn: Got you, let's leave it as is.

alxkzmn (approved):
LGTM!

brech1 (approved):
LGTM 🚀

PR #231, opened 2026-02-11
fix: update GITHUB_TOKEN to use fallback to github.token

alxkzmn:


alxkzmn:
Will merge as the test run **without** the GH_PAT in my fork was successful: https://github.com/alxkzmn/csp-benchmarks/pull/3

PR #236, opened 2026-02-13
chore(ligetron): update submodule to v1.4.0 and adjust build setup

jdodinh:
- Update ligero-prover submodule from v1.0.0 to v1.4.0
- Adjust build to fetch WABT 1.0.39 and cache it with Dawn
- Update Dawn commit to cec4482
- Fix SDK path (sdk → sdk/cpp)
- Update proof output path (proof.data → proof_data.gz)

### DIFF
--- a/ligetron/osx_local_setup.sh
+++ b/ligetron/osx_local_setup.sh
@@ -120,25 +120,13 @@ ok "emsdk ready (emcmake available)"
 # Ligetron submodule
 # -----------------------
 LIGETRON_DIR="${SCRIPT_DIR}/ligero-prover"
-LIGETRON_VERSION="1.4.0"
 
 if [[ ! -d "${LIGETRON_DIR}" ]]; then
   step "Ligetron submodule not found; initializing"
   git -C "${REPO_ROOT}" submodule update --init --recursive ligetron/ligero-prover || {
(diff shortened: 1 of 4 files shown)


brech1:
Is there a particular reason to pin wabt on the ci and build from source? brew is correctly installing it on latest actions, see:

- https://github.com/privacy-ethereum/csp-benchmarks/actions/runs/21953983579/job/63412580660

If it's strictly necessary version could also be pinned with brew.

And regarding `osx_install_script`, the submodule version is already being enforced, the check seems redundant imo. 

Don't have a strong opinion on these though, just comments. Rest of PR LGTM, just small whitespace nit.

ligetron/osx_local_setup.sh
```diff
@@ -24,6 +24,7 @@ if [[ "$(uname -s)" != "Darwin" ]]; then
   echo "This script is for macOS (Darwin) only."; exit 1
 fi
 
+
```
brech1: nit

jdodinh:
@brech1 You're right about wabt, we can just use the brew version, it seems to work fine. I had issues with 1.0.39 earlier, but can't reproduce them anymore...

Also, removing the redundant submodule version check.

alxkzmn (approved):
LGTM!

alxkzmn (approved):

brech1 (approved):

Issue #207, opened 2026-01-20
Implement Plonky2 Keccak benchmark


alxkzmn:
https://github.com/qope/plonky2-keccak256

AnInsaneJimJam:
Can I take this Issue?

alxkzmn:
@AnInsaneJimJam Sure! We have a deadline of this Friday (Feb 20th) to prepare all circuits to run the next round of benchmarks, would be nice to have it by then.

AnInsaneJimJam:
Sure. Thank you !

AnInsaneJimJam:
@alxkzmn I have added the PR. It would be great if you can review it. #238

PR #238, opened 2026-02-20
feat(plonky2) : Add Keccak benchmarks

AnInsaneJimJam:
This pr adds keccak benchmarks to the plonky2 proving system (rust based)
- Added a folder keccak256 in src
- It contains the circuit.rs (main circuit logic) and u64target.rs ( field to u64 logic). Both files are taken from https://github.com/qope/plonky2-keccak256

This pr also improves the folder system in the plonky2.
- Earlier there was only sha256 files, so they were directly in src. Although there was poseidon( but it didn't need a circuit.rs as it is imported directly from plonk2)
- Now src contains 2 folders - one each for keccak256 and sha256 with their relevant files
- Also earlier package name was defined as plonky2_sha256 but since it's includes both. It has been renamed to plonky2_circuits.

This pr is correcponding to issue #207 .

Below is the screenshot of benchmarks wih reduced flag set
Screenshot from 2026-02-20 17-23-41


plonky2/benches/keccak.rs
```diff
@@ -0,0 +1,43 @@
...
+            .to_bytes(&gate_serializer)
+            .unwrap()
+            .len();
+        let generator_serializer = U32GeneratorSerializer::<C, D>::default();
+        let prover_data_size = circuit_data
+            .prover_only
+            .to_bytes(&generator_serializer, &circuit_data.common)
+            .unwrap()
+            .len();
+        prover_data_size + common_data_size
```
alxkzmn: This code is identical to the SHA-256 case - can be extracted into a function and reused across both benchmarks

plonky2/src/bin/keccak_mem.rs
```diff
@@ -0,0 +1,20 @@
...
+struct Args {
+    /// Input size parameter
+    #[arg(long)]
+    input_size: usize,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    keccak_no_lookup_mem(args.input_size);
```
alxkzmn: No need to keep "no_lookup" in the name as we're not trying to distinguish between different circuits by whether they're using lookup or not.

alxkzmn:
Please include the original license (https://github.com/qope/plonky2-keccak256/blob/main/MIT-LICENSE.txt) into `./plonky2/src/keccak256` folder.

brech1:
Thank you for this PR! Don't have any comments other than I agree that if the circuit code belongs to that repository it should be correctly attributed.

plonky2/src/bin/keccak_mem.rs
AnInsaneJimJam: What should be the new name then keccak_mem and sha256_mem ?

AnInsaneJimJam:
@brech1 Why did you force push my changes on top of the ligetron scripts? Just Curious? Any particular reason

AnInsaneJimJam:
@alxkzmn @brech1 I have done the requested changes. Can you review it once?

alxkzmn (approved):
LGTM!

brech1:
> @brech1 Why did you force push my changes on top of the ligetron scripts? Just Curious? Any particular reason

I just rebased the branch to latest `main` to trigger the workflows with proper changes :)

AnInsaneJimJam:
> > @brech1 Why did you force push my changes on top of the ligetron scripts? Just Curious? Any particular reason
> 
> I just rebased the branch to latest `main` to trigger the workflows with proper changes :)

Got it! Thanks !

brech1 (approved):
LGTM!

PR #247, opened 2026-02-24
fix: risc0 ci workflow and guest versions

brech1:
# Description

Update risc0 versions to `v3.0.5` in both CI and guests.


alxkzmn (approved):
LGTM!

PR #248, opened 2026-02-25
chore(ligetron): adjust ligero-prover version

jdodinh:
@alxkzmn As discussed in private, the Ligero team renamed v1.4.0 to v1.3.1. This doesn't change the code but makes the submodule commit match the relevant version in the origin. Version updates will be smoother in the future.


alxkzmn (approved):
LGTM!

PR #264, opened 2026-03-18
Adjust Jolt configuration for 2048 B input size of SHA-256

alxkzmn:
The max trace length of 65536 was insufficient for running the SHA-256 circuit benchmark with 2048 B input (see error log:
https://github.com/privacy-ethereum/csp-benchmarks/actions/runs/23023028374/job/66948508069). We set the max trace size for 2kB input to 131072 to accommodate.


alxkzmn:
cc @0xAndoroid

PR #272, opened 2026-04-09
chore(noir): update noir version and use optimized poseidon2 library

TomAFrench:
This PR bundles 3 changes to the noir benchmarks:

- They're now run under nargo 1.0.0-beta.19 and Barretenberg v5.0.0-nightly.20260324.
- The `install-nargo` action now delegates to the official `noirup` action (I've maintained the `install-nargo` action for consistency but it can be removed)
-  The  `poseidon2` benchmark has been updated to use the official [poseidon library](https://github.com/noir-lang/poseidon) which exposes both poseidon and poseidon2. I've swapped this out as the official library makes use of a more optimized implementation of poseidon2.

### DIFF
--- a/.github/actions/install-nargo/action.yml
+++ b/.github/actions/install-nargo/action.yml
@@ -5,14 +5,12 @@ inputs:
   version:
     description: Version of Nargo/Noir binaries.
     required: false
-    default: "1.0.0-beta.13"
+    default: "1.0.0-beta.19"
 
 runs:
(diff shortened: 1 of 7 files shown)


alxkzmn (changes requested):
Thanks a lot for the contributions! I'll fix a couple more things and merge.

alxkzmn (approved):

PR #284, opened 2026-05-11
Add ProveKit-Groth16 benchmark

Bisht13:
Adds `provekit-groth16` alongside `provekit/` (Spartan+WHIR), benchmarking the **Groth16+BSB22** flavour of [ProveKit](https://github.com/worldfnd/ProveKit) on the standard targets (sha256, ecdsa, keccak, poseidon, poseidon2).

Trusted setup is sampled per `prepare` call (toxic waste zeroized on drop) — benchmark-only; production needs a ceremony.


provekit-groth16/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        .expect("Groth16 proof generation failed")
+}
+
+pub fn verify(prepared: &GrothPrepared, proof: &NoirProof) -> Result<(), &'static str> {
+    let mut verifier = prepared.verifier.clone();
+    verifier.verify(proof).map_err(|_| "Proof is not valid")
+}
+
+pub fn preprocessing_size(prepared: &GrothPrepared) -> usize {
+    prepared.pk_size
```
alxkzmn: By preprocessing artifacts we actually mean everything that a prover needs to generate the proof, so in this case it should also be the R1CS, witness generator binaries and anything else that a deployed application would include.

alxkzmn (changes requested):
Please account for all prover artifacts to calculate the preprocessing size.

provekit-groth16/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        toml_path,
+        pk_size,
+        num_constraints,
+    }
+}
+
+pub fn prepare_sha256(input_size: usize) -> GrothPrepared {
+    let package = format!("csp_sha256_{input_size}");
+    let (scheme, _) = compile_package(&package);
+    let (data, _digest) = utils::generate_sha256_input(input_size);
```
alxkzmn: Why the digest isn't used?

provekit-groth16/src/lib.rs
Bisht13: Reads the PKP size now.

provekit-groth16/src/lib.rs
Bisht13: The sha256 circuits use -> pub [u8; 32] (e.g. circuits/sha256_128/src/main.nr:5), so the digest is a public output of the proof, not a Prover.toml input. Same handling as the existing provekit/crate (provekit/src/lib.rs:97).

alxkzmn: I think we need to either check the public output, or provide it as an input and `assert(digest == expected)`

Bisht13: Updated the sha256 circuits to take result: [u8; 32] and assert(digest == result), matching keccak's pattern. prepare_sha256 now writes the expected digest to Prover.toml.

alxkzmn (approved):
LGTM!

Issue #132, opened 2025-10-31
Implement Circom/Groth16 ECDSA benchmarks


alxkzmn:
Research further into which ECDSA implementations we want to include. If we include multiple, we use the "feat" property to distinguish them:
       - https://github.com/0xPARC/circom-ecdsa (for secp256k1 curve) (Crema labs fork may have better circomlib deps management: https://github.com/crema-labs/circom-ecdsa)
       - https://github.com/privacy-ethereum/circom-ecdsa-p256 (for p256 curve)
       - https://github.com/crema-labs/ecdsa-p384-circom (for p384 curve)

brech1:
Even if we just include one we need the curve differentiator imo

alxkzmn:
It could be the feature field.

Spyro7883:
Is this still wanted? I'd like to take it, starting with secp256k1 and putting the curve in the feat field, matching the risc0/jolt entries.

alxkzmn:
@Spyro7883 yes, and please use the latest results from https://zk.golf/challenges/secp256k1-scalar-mul and https://zk.golf/challenges/secp256k1-fixed-base-scalar-mul to implement the circuit efficiently.

Spyro7883:
Progress update: ECDSA verification is at **503,280 nonlinear constraints** (532,811 total), against **1,508,904** for `ECDSAVerifyNoPubkeyCheck(64, 4)`. Both linked challenges are in it: 4-dimensional fake-GLV for the variable base (417,661 alone, from 1,399,441) and a width-12 comb for the fixed base (65,601, from 95,444). Semantics follow k256's `verify_prehash`, so the baseline is doing strictly less work. Branch: [`Spyro7883/csp-benchmarks:circom-ecdsa-glv4`](https://github.com/Spyro7883/csp-benchmarks/tree/circom-ecdsa-glv4).

One finding worth flagging: Straus over `Secp256k1AddUnequal` is forgeable on coincident operands, not merely incomplete, so the loop uses a strict addition plus two guards. Witness tests pass 21/21, including that case.

Three questions before I open the PR:

1. **I can't upload the zkey myself.** It's 309 MB and lives in the `zkeys-v1` release, which only maintainers can write to. #191 and #200 could still ship theirs through LFS, but #279 moved them. Can you upload it? If it helps, I can make `ZKEYS_RELEASE_URL` fall back to the PR's head repository, so I can host mine until you publish yours.

[...]

PR #302, opened 2026-08-22
Add secp256k1 ECDSA verification benchmark for circom

Spyro7883:
Part of #132: secp256k1.

Adds a circom ECDSA verification benchmark over secp256k1, registered through the shared harness with the curve in the feature field. The circuit verifies a prehashed signature with the same semantics as `k256::verify_prehash`, and takes the digest, the signature and the public key - same interface as the other ECDSA circuits here.

It applies both zk.golf techniques: a 4-dimensional fake-GLV for the variable base ([secp256k1-scalar-mul](https://zk.golf/challenges/secp256k1-scalar-mul)) and a width-12 comb for the fixed base ([secp256k1-fixed-base-scalar-mul](https://zk.golf/challenges/secp256k1-fixed-base-scalar-mul)).
Together they bring verification to 503,280 nonlinear constraints, down from 1,508,904 for `ECDSAVerifyNoPubkeyCheck(64, 4)`.

Building the crate now needs `circom` on PATH: the witness generator for this circuit is compiled from the circuit rather than committed, and CI installs the pinned compiler to do it. Every other circuit still ships its `.cpp` and `.dat`.

The `.zkey` isn't included, as agreed in #132, so the benchmark job stays red until it's uploaded.


alxkzmn:
@Spyro7883 please fix the missing circomlib issue (see the CI errors - the code should point to the correct circomlib location inside https://github.com/ethereum/csp-benchmarks/tree/main/circom)

Spyro7883:
The failure is caused by the absence of `ecdsa_32_0001.zkey` in the `zkeys-v1` release. The key needs generating with `powersOfTau28_hez_final_20.ptau`, not `_19`: snarkjs counts 532,811 constraints, which is over 2^19.

circomlib is fixed in `dce45cc` and `bbb45a3`.

Spyro7883:
circom/checksums.sha256 has the hash of the key I built locally, so the verify step will fail until that line matches yours. Send me the hash and I'll update it, or you can push the line if that's faster.

Spyro7883:
Read through 1942c5c9 - the equal-x split is the one I got wrong; rejecting R = -[u1]G rejects valid signatures.

I've dropped the ECDSA zkey line from rust_benchmarks_parallel.yml and its line from checksums.sha256, since no key is published yet and circom-ecdsa-bench already gates the ECDSA benchmark off by default. Both can be restored once a key is published.

I recompiled with --O2 locally and get 512,955 constraints, so powersOfTau28_hez_final_19.ptau is enough now, not _20 as I said.

alxkzmn:
@Spyro7883 the PR looks good, thanks for your work! The CI failure is unrelated, I'll fix it separately.

alxkzmn (approved):
LGTM!

Issue #128, opened 2025-10-29
Add Halo2 SHA-256 benchmarks


alxkzmn:

AnInsaneJimJam:
Hey @alxkzmn Can I take this issue?

alxkzmn:
@AnInsaneJimJam sure, I'd suggest you to use Axiom's Halo2 with their zkevm-hashes circuits. I wouldn't mind if you add more workloads than just SHA-256 (e.g., Keccak), if they are well-optimized.

AnInsaneJimJam:
on it !!

AnInsaneJimJam:
https://github.com/axiom-crypto/halo2-lib/tree/develop/hashes/zkevm. Should I add these two circuits ? These lgtm.

AnInsaneJimJam:
Hey @alxkzmn . Added a PR for this: #304

PR #304, opened 2026-09-06
Add Halo-2 benchmarks

AnInsaneJimJam:
This PR solves issue #128 by integrating Axiom's Halo2 Circuits for Sha256 and Keccak. 
Here is the implementation: https://github.com/axiom-crypto/halo2-lib/tree/develop/hashes/zkevm.
Three magic numbers are inherited along with the circuit code from the upstream:

- BLINDING_ROWS=109 . Rows withheld from the end of the domain so witness assignment never collides with halo2's blinding factors. usable_rows(k) = 2^k - 109. No effect on bechmnarks as even if we decrease it to 9, the K remains same

- MIN_K = 10 (superficial) : Mathematically verified.

- KECCAK_ROWS_PER_ROUND = 28. Upstream test 4 values:  28, 25, 9, and 5. Didn't dig into which one will be better.


AnInsaneJimJam:
Actually 28 is the fastest among all.

alxkzmn (approved):
LGTM! Thanks @AnInsaneJimJam !

